PR #18, opened 2020-07-23
Fix grammers-client compilation when using it as git dependency

Manuel1510:


Lonami:
The problem with this is that I'm pretty sure you can't publish to crates.io with path dependencies. While it would work while developing and using from `git` or `path` I do not think this is the right solution unfortunately.

Manuel1510:
According to this, you can use both a path/git location and a registry version, the path would be used when building locally and crates.io would ignore it: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

Lonami:
Ah, I see the way that it works now. That's reasonable enough.

PR #23, opened 2020-10-13
Refactor client chat resolve_username

pickfire:


Lonami:
Yeah, that's a lot cleaner, thanks!

PR #24, opened 2020-11-09
Use read_buf for reading

Selicre:
This uses `read_buf` for reading from the network, removing the need for transport to calculate the missing bytes.

### DIFF
--- a/lib/grammers-mtsender/src/lib.rs
+++ b/lib/grammers-mtsender/src/lib.rs
@@ -44,7 +44,6 @@ pub struct Sender<T: Transport, M: Mtp> {
 
     // Transport-level buffers and positions
     read_buffer: Vec<u8>,
-    read_index: usize,
     write_buffer: Vec<u8>,
     write_index: usize,
 }
@@ -75,7 +74,6 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
             requests: vec![],
 
             read_buffer: Vec::with_capacity(MAXIMUM_DATA),
-            read_index: 0,
             write_buffer: Vec::with_capacity(MAXIMUM_DATA),
             write_index: 0,
         })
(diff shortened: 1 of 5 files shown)


lib/grammers-mtsender/src/lib.rs
```diff
@@ -156,19 +152,18 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
             //
             // Note that this mutably borrows `self`, so the caller can't `enqueue` other requests
             // while reading from the network, which means there's no need to handle that case.
-            trace!("reading up to {} bytes from the network", read_len);
+            trace!("reading up to ∞ bytes from the network");
```
Lonami, line 155:
```suggestion
            trace!("reading bytes from the network");
```

lib/grammers-mtsender/src/lib.rs
```diff
@@ -156,19 +152,18 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
...
-                .read(&mut self.read_buffer[self.read_index..])
+                .read_buf(&mut self.read_buffer)
                 .await?;
 
             self.on_net_read(n)
         } else {
             trace!(
-                "reading up to {} bytes and sending up to {} bytes via network",
-                read_len,
+                "reading up to ∞ bytes and sending up to {} bytes via network",
```
Lonami, line 163:
```suggestion
                "reading bytes and sending up to {} bytes via network",
```

Lonami (changes requested):

lib/grammers-mtsender/src/lib.rs
```diff
@@ -276,13 +251,9 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
         {
             Ok(_) => {
                 self.read_buffer.clear();
-                self.read_index = 0;
                 self.process_mtp_buffer().map_err(|e| e.into())
             }
             Err(transport::Error::MissingBytes(n)) => {
```
Lonami, line 256:
```suggestion
            Err(transport::Error::MissingBytes(_)) => {
```

I wonder if it's worth keeping the amount `n` at all. If we don't ignore it, perhaps we could call `.reserve()` instead? But if the buffer already has enough size it's pointless. Anyway as is this is a warning I'm pretty sure.

Selicre: I was thinking of entirely removing the calculation of it; I think `.reserve()` is already called in the most optimal way regardless

PR #26, opened 2020-11-10
[WIP] grammers-client: various friendly functions and improvements

rupansh:
add
```
client.edit_message
client.get_reply_message
mesage.reply_to_message_id
message.from_id
```
Also make SignInError public


Lonami:
Merging this as-is so that I can work from there.

PR #27, opened 2020-11-11
Use BytesMut for buffers instead of Vec

Selicre:
This allows for further architecture improvements in the future, and simplifies transport code.


Lonami:
Failure seems unrelated to this PR..

PR #29, opened 2020-11-11
Calculate the first salt from authentication

habnabit:
`new_nonce ^ server_nonce` is used as the first salt. Calculating this
locally means the first message sent after a new connection won't be
rejected for having an incorrect salt and have to be retried.

Instead of making `create_key` return a 3-tuple to include the new
data, there's a new struct `Finished` to bundle all the data together.


Lonami:
Failure seems not related to this PR (did they just update their library? I'm pretty sure this wasn't an issue before).

PR #31, opened 2020-11-19
Change dialogs example info line

DragSama:


Lonami:
Good catch, but I think I'll add the change along with the rest required to fix the build. I'm not a fan of these tiny commits.

PR #32, opened 2020-11-28
Support two-factor authentification

AlexPl292:
Changes:
- Updating sha2 library. Needed by `pbkdf2` since we don't want to reimplement this algorithm
- Implementing 2fa


lib/grammers-client/examples/dialogs.rs
```diff
@@ -58,7 +61,19 @@ async fn async_main() -> Result<()> {
...
         let code = prompt("Enter the code you received: ")?;
-        client.sign_in(&token, &code).await?;
+        let signed_in = client.sign_in(&token, &code).await;
+        match signed_in {
+            Ok(_) => (),
+            Err(SignInError::PasswordRequired) => {
+                let password = prompt("Enter the password: ")?.to_bytes();
+
+                // Clean extra symbols (no idea where they come from)
+                let password = password[1..(password.len() - 2)].to_vec();
```
Lonami, line 71: This needs cleaning up.

lib/grammers-client/src/client/auth.rs
```diff
@@ -261,6 +265,58 @@ impl Client {
...
+            Err(InvocationError::Rpc(RpcError { name, .. }))
+                if name.starts_with("SESSION_PASSWORD_NEEDED") =>
+            {
+                Err(SignInError::PasswordRequired)
+            }
+            Err(error) => Err(SignInError::Other(error)),
+        }
+    }
+
+    pub async fn two_factor_auth(
```
Lonami, line 277: We may want to bikeshed on the naming here, and also we have to document this method with usage examples.

lib/grammers-client/src/client/auth.rs
```diff
@@ -261,6 +265,58 @@ impl Client {
...
+        &mut self,
+        password: Vec<u8>,
+    ) -> Result<tl::types::User, SignInError> {
+        let request = tl::functions::account::GetPassword {};
```
Lonami, line 281: We might want to perform this call when we get `SESSION_PASSWORD_NEEDED`, so that the `SignInError::PasswordRequired` can include the hint (that the caller can display somehow) along with a "token" (similar to how send code request gives you one for sign in) that has to be used to call this method.

lib/grammers-client/examples/dialogs.rs
```diff
@@ -58,7 +61,19 @@ async fn async_main() -> Result<()> {
         let phone = prompt("Enter your phone number (international format): ")?;
         let token = client.request_login_code(&phone, api_id, &api_hash).await?;
         let code = prompt("Enter the code you received: ")?;
-        client.sign_in(&token, &code).await?;
+        let signed_in = client.sign_in(&token, &code).await;
+        match signed_in {
+            Ok(_) => (),
+            Err(SignInError::PasswordRequired) => {
+                let password = prompt("Enter the password: ")?.to_bytes();
```
Lonami, line 68: This is a good showcase on how to sign in when password is required, but we probably want to add a comment saying something along the lines of:

```rust
// Note: this `prompt` method will echo the password in the console.
//       Real code might want to use a better way to handle this.
```

lib/grammers-client/src/client/auth.rs
```diff
@@ -261,6 +265,58 @@ impl Client {
...
+                m1,
+            }),
+        };
+
+        match self.invoke(&check_password).await {
+            Ok(tl::enums::auth::Authorization::Authorization(x)) => {
+                // Safe to unwrap, Telegram won't send `UserEmpty` here.
+                Ok(x.user.try_into().unwrap())
+            }
+            Ok(tl::enums::auth::Authorization::SignUpRequired(x)) => {
```
Lonami, line 315: I do not think Telegram can send a sign up required here. I think we should panic if this variant is returned (kind of "unreachable", as it's done in other places; but not an actual unreachable because we never know what Telegram might do).

lib/grammers-client/src/client/auth.rs
```diff
@@ -261,6 +265,58 @@ impl Client {
...
+                Err(SignInError::SignUpRequired {
+                    terms_of_service: x.terms_of_service.map(|tos| tos.into()),
+                })
+            }
             Err(error) => Err(SignInError::Other(error)),
```
Lonami, line 320: We might want add a new variant for "invalid password" to let the user know more nicely.

lib/grammers-crypto/src/two_factor_auth.rs
```diff
@@ -0,0 +1,170 @@
...
+            let mut hasher = Sha256::new();
+            $(
+                hasher.update($x);
+            )*
+            hasher.finalize()
+        }
+    };
+}
+
+pub fn calculate_2fa(
```
Lonami, line 26: I'd like it if this method's documentation pointed out where this implementation comes from (presumably, a link to https://core.telegram.org/api/srp, if that's what you used).

lib/grammers-crypto/src/two_factor_auth.rs
```diff
@@ -0,0 +1,170 @@
...
+        .map(|(&x1, &x2)| x1 ^ x2)
+        .collect();
+}
+
+fn pad_to_256(data: &Vec<u8>) -> Vec<u8> {
+    let mut new_vec = data.clone();
+    for _ in 0..(256 - new_vec.len()) {
+        new_vec.insert(0, 0);
+    }
+    new_vec
```
Lonami, line 132: Repeatedly inserting at the front has to move data around a lot. I think it would be cleaner to do the following:

```suggestion
    let mut new_vec = vec![0; 256 - data.len()];
    new_vec.extend(data);
    new_vec
```

Lonami (changes requested):

lib/grammers-client/src/client/auth.rs
AlexPl292: I don't know about another name :/ For me, this one looks nice, but we can use `password_check` for example?

AlexPl292:
Hey, here are the updates for the PR. I did a separate commit, but I can merge it into the previous if you like.

lib/grammers-client/src/client/auth.rs
```diff
@@ -251,20 +281,128 @@ impl Client {
...
+    ///         panic!("Failed to sign in as a user :(\n{}");
+    ///     }
+    /// };
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn two_factor_auth(
+        &mut self,
+        password: Vec<u8>,
+        password_information: tl::types::account::Password,
```
Lonami, line 370: The signature of `Client::sign_in` is `(token: &LoginToken, code: &str)`. I think it would make sense if the signature here was the same, that is, first the "token" and then the password itself.

I am also not to keen on requiring a `Vec<u8>` as input, since most users will have "normal" passwords that don't need anything special (I am not aware of any official client that allows you to use arbitrary bytes as your password, as it would be pretty impractical). Maybe we can take `AsRef<[u8]>` instead?

lib/grammers-client/src/client/auth.rs
```diff
@@ -14,25 +14,51 @@ use grammers_tl_types as tl;
...
+
 /// The error type which is returned when signing in fails.
 #[derive(Debug)]
 pub enum SignInError {
-    SignUpRequired {
-        terms_of_service: Option<tl::types::help::TermsOfService>,
-    },
     InvalidCode,
+    IncorrectParameters(VerificationError),
+    IncorrectPassword,
```
Lonami, line 24: Above we're using `Invalid` instead of `Incorrect` code. Perhaps it would make sense to stick to the same naming convention?:

```suggestion
    InvalidParameters(VerificationError),
    InvalidPassword,
```

lib/grammers-client/src/client/auth.rs
```diff
@@ -14,25 +14,51 @@ use grammers_tl_types as tl;
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum SignInResult {
+    SignedIn(tl::types::User),
+    PasswordRequired {
+        password_information: tl::types::account::Password,
+    },
```
Lonami, line 45: I think I would prefer this to be something like the following:

```suggestion
    PasswordRequired(PasswordToken),
```

This is more in line with how `send_code_request` works (it returns an opaque "token" you can later use), and the hint can be trivially accessible with:

```rust
impl PasswordToken {
    fn hint(&self) -> Option<&str> {
        self.password.hint.as_ref()
    }
}
```

The password token simply contains the password as a private field (we want to avoid exposing TL types where possible), even if not done yet for the TOS.

lib/grammers-client/src/client/auth.rs
```diff
@@ -240,7 +270,7 @@ impl Client {
         &mut self,
         token: &LoginToken,
         code: &str,
-    ) -> Result<tl::types::User, SignInError> {
+    ) -> Result<SignInResult, SignInError> {
```
Lonami, line 273: I am not too sure on this. I would argue requiring a password or signing up are "errors" because sign in itself failed. Forcing the user to handle the "error" case even when it returns `Ok` (as in, "hey, sign in failed, you actually need to sign up") doesn't feel right.

So I would argue sign up required / password needed should be part of sign in error.

lib/grammers-client/examples/dialogs.rs
```diff
@@ -58,7 +60,29 @@ async fn async_main() -> Result<()> {
...
+                // Note: this `prompt` method will echo the password in the console.
+                //       Real code might want to use a better way to handle this.
+                let hint = password_information.hint.as_ref().unwrap();
+                let prompt_message = format!("Enter the password (hint {}): ", &hint);
+                let mut password = prompt(prompt_message.as_str())?.into_bytes();
+
+                // Remove \n symbol if any
+                if password.len() > 0 && password.last().unwrap() == &10u8 {
+                    password.remove(password.len() - 1);
+                }
```
Lonami, line 77: Can this actually happen? Would a simple [`str::trim`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim) work here, for the sake of the example? I don't think many users are out there with leading or trailing whitespace in their passwords. I don't even think official applications let you enter a new line in a password. Is the newline something our `prompt` returns?

lib/grammers-client/src/client/auth.rs
```diff
@@ -251,20 +281,128 @@ impl Client {
...
             Err(InvocationError::Rpc(RpcError { name, .. })) if name.starts_with("PHONE_CODE_") => {
                 Err(SignInError::InvalidCode)
             }
             Err(error) => Err(SignInError::Other(error)),
         }
     }
 
+    /// Extract information needed for the two-factor authentication
+    /// It's called automatically when we get SESSION_PASSWORD_NEEDED error during sign in.
+    pub async fn get_password_information(
```
Lonami, line 308: Do we want this method to be public? Why is the `Err` variant a `SignInError`, shouldn't it simply be `InvocationError`?

lib/grammers-client/src/client/auth.rs
```diff
@@ -251,20 +281,128 @@ impl Client {
...
+        let data: tl::enums::account::Password = match self.invoke(&request).await {
+            Ok(x) => x.into(),
+            Err(error) => return Err(SignInError::Other(error)),
+        };
+
+        let password_information: tl::types::account::Password = match data {
+            tl::enums::account::Password::Password(pass) => pass,
+        };
+
+        Ok(password_information)
```
Lonami, line 322: After changing the return error, this should probably be:

```suggestion
        let password = self.invoke(&request).await?.into();
        Ok(password)
```

Lonami (changes requested):

lib/grammers-client/src/client/auth.rs
AlexPl292: Okay, I agree about "sign up required", but I think "password needed" can be ok result. If we consider `signIn` as a separate function, then yes, it's an error, sign in failed. But if we look at the whole process, this is a fine result "hey, your token code is okay, but we need a password now".

However, I'm okay to move it to errors, this looks like a matter of taste after all. And telegram considers it as an error. Do you want to move it to the error?

lib/grammers-client/src/client/auth.rs
AlexPl292: Agree about `SignInError`.

This function is public because I wanted to provide a way for case telegram sends incorrect data with a `password` object. There is a verification for `p` and `g` parameters and docs say that we should perform it. I don't think that telegram can send incorrect values, but still...
Okay, another opinion. I make this function public and call it again if the parameters are incorrect to get the new params. We can call it for the new params two times more and panic with `telegram is broken, go use whatapp` (kidding) in case tg will continue to send incorrect params. What do you think?

lib/grammers-client/src/client/auth.rs
Lonami: Yes, let's move it to errors, along with the sign up required.

lib/grammers-client/src/client/auth.rs
Lonami: Ah yeah I forgot to mention this earlier. If the parameters are invalid, I think the library should retry in-place once, and if that fails again, then we return invocation failed. Users can't really do much about this, and fetching the required parameters should probably not be responsibility of the user (why should they care how 2FA is handled?). Then this method can become private.

AlexPl292: > then we return invocation failed.

The method returns `SignInError`. Invocation error can be wrapped by `SignInError::Other`, but it doesn't seem nice to return `SESSION_PASSWORD_NEEDED`. Looks like we need either 1) Create a new invocation error with `failed` or 2) Create a new `SignInError` or 3) Panic. TBH, none of this approaches looks good to me. What do you think about it?

Lonami: Let's `panic!()`, since this might indicate a bug somewhere.

lib/grammers-client/src/client/auth.rs
AlexPl292: I've named it `check_password` for now. This is how tg calls it. And looks like we've got no better variants in tg chat :/

lib/grammers-client/src/client/auth.rs
```diff
@@ -7,20 +7,25 @@
...
+
 /// The error type which is returned when signing in fails.
 #[derive(Debug)]
 pub enum SignInError {
     SignUpRequired {
         terms_of_service: Option<tl::types::help::TermsOfService>,
     },
+    PasswordRequired(PasswordToken),
     InvalidCode,
+    InvalidParameters,
```
Lonami, line 27: We've decided this variant won't be exposed to the user and we'll panic instead so there's no need to have that here.

lib/grammers-client/src/client/auth.rs
```diff
@@ -258,13 +270,129 @@ impl Client {
...
+        let (_, _, g, p) =
+            Client::extract_password_parameters(password.current_algo.as_ref().unwrap());
+
+        // Telegram sent us incorrect parameters, trying to get them again
+        if !check_p_and_g(p, g) {
+            password = self.invoke(&request).await?.into();
+            let (_, _, g, p) =
+                Client::extract_password_parameters(password.current_algo.as_ref().unwrap());
+            if !check_p_and_g(p, g) {
+                panic!("Cannot get correct password information from telegram")
```
Lonami, line 305:
```suggestion
                panic!("Failed to get correct password information from Telegram");
```

lib/grammers-client/src/client/auth.rs
```diff
@@ -258,13 +270,129 @@ impl Client {
...
             Err(InvocationError::Rpc(RpcError { name, .. })) if name.starts_with("PHONE_CODE_") => {
                 Err(SignInError::InvalidCode)
             }
             Err(error) => Err(SignInError::Other(error)),
         }
     }
 
+    /// Extract information needed for the two-factor authentication
+    /// It's called automatically when we get SESSION_PASSWORD_NEEDED error during sign in.
+    async fn get_password_information(&mut self) -> Result<PasswordToken, InvocationError> {
```
Lonami, line 291: I was thinking, if the parameters are time-sensitive, we might want to re-fetch them inside `check_password` (because then we'll use them immediately), not twice in `sign_in` (user may take a while to input the password).

Lonami (changes requested):
Sorry for so much back and forth but I'd like to merge this on its best possible state (almost there) :P

lib/grammers-client/src/client/auth.rs
AlexPl292: Sorry, just forgot to remove this one

lib/grammers-client/src/client/auth.rs
AlexPl292: You mean, we fetch the parameters in `sign_in`, then ask the user for a password, check the params in `check_password` and refetch (if necessary) in `check_password` as well? This might work, why not.

AlexPl292:
> Sorry for so much back and forth but I'd like to merge this on its best possible state (almost there) :P

It's absolutely okay, don't worry. Let's make it so that I won't feel the shame when I'll see this code a year later :)

lib/grammers-client/src/client/auth.rs
AlexPl292: However, did an experiment with a two minutes delay to enter the password and it worked fine.
TBH, I would say that it would be cool if we could verify the params in the background while the user enters his password (because it takes some time).

AlexPl292:
In addition, I want to check the performance of the operations and maybe compare it with other solutions in other languages. Maybe there is a room for optimization to speed it up a bit because currently there is a delay in about 5-8 seconds for params verification and M1 parameter calculation. Not sure if this should be in this PR or not.

lib/grammers-client/src/client/auth.rs
Lonami: > You mean, we fetch the parameters in `sign_in`, then ask the user for a password, check the params in `check_password` and refetch (if necessary) in `check_password` as well?

Yes, exactly that.

> did an experiment with a two minutes delay to enter the password and it worked fine.

Yeah but who knows when working with Telegram, perhaps it randomly fails at times to make sure clients verify it and two attempts is enough. Or perhaps Telegram will never fail and it's only to prevent malicious attackers. Or perhaps in re-fetch it gives fresh values (haven't tried) so it's okay to do anyway.

> it would be cool if we could verify the params in the background while the user enters his password

I am not concerned with speed at the moment, this would be a fair bit of complexity not worth it right now. Furthermore, if the user knows their password, there's no need to ask for input as they may hardcode it and just call the method immediately. Then they can do this with as many clients or tasks they want to make the most of their CPUs.

Lonami:
> Not sure if this should be in this PR or not.

Up to you, this looks good to me so we can merge it now and focus on performance improvements later. We could even add a `client.sign_in_with_password` later on for when the user knows they have a password (not in this PR).

Lonami (approved):

AlexPl292:
Should be good so far, if I didn't miss something.

AlexPl292:
JFI regarding the performance: nevermind, it's not 5 sec, it's ~90ms. Just don't be an idiot like me and don't forget to enable --release :)

PR #35, opened 2020-12-01
grammers-client: files: add upload_stream

rupansh:


lib/grammers-client/src/client/files.rs
```diff
@@ -152,6 +153,105 @@ impl ClientHandle {
...
+    ///
+    /// let uploaded_file = client.upload_stream(stream, Some("sleep.jpg".to_string())).await?;
+    ///
+    /// client.send_message(&chat, InputMessage::text("Check this out!").photo(uploaded_file)).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// [`InputMessage`]: crate::InputMessage
+
```
Lonami, line 177:
```suggestion
```

lib/grammers-client/src/client/files.rs
```diff
@@ -152,6 +153,105 @@ impl ClientHandle {
...
+    pub async fn upload_stream<S: AsyncRead + AsyncSeek + Unpin>(
+        // Do we want Sized instead of AsyncSeek?
```
Lonami, line 179: `Sized` won't tell you the size of the file though? I don't see the purpose of this comment.

I think the right approach here is to take the file size as input parameter, and read that much from the stream. Think HTTP streams, may not be seekable, but you might know the file size beforehand.

lib/grammers-client/src/client/files.rs
```diff
@@ -152,6 +153,105 @@ impl ClientHandle {
...
+        &mut self,
+        stream: &mut S,
+        name: Option<String>, // Cries about type hint if None is passed if we use Into<String> instead
```
Lonami, line 182: This can probably be something with `AsRef`, maybe `Option<AsRef<str>>` works, but in any case remove that comment:

```suggestion
        name: Option<String>,
```

Lonami (changes requested):
We likely want to reimplement `upload_file` to use `upload_stream` instead. It's never good to have two implementations of essentially the same function.

Lonami (changes requested):
Didn't we settle on not requiring `AsyncSeek` and instead read in a `size: usize`?

lib/grammers-client/src/client/files.rs
```diff
@@ -138,21 +141,21 @@ impl ClientHandle {
...
         {
-            file.write_all(&chunk)?;
+            file.write_all(&chunk).await?;
         }
 
         Ok(())
     }
 
-    /// Uploads a local file to Telegram servers.
+    /// Uploads an async buffer to Telegram servers.
```
Lonami, line 158:
```suggestion
    /// Uploads an async stream to Telegram servers.
```

Lonami (approved):

PR #36, opened 2020-12-19
entity parser: update to new naming

sabbyX:
I think these are missed in ab05f96, Wonder how CI doesn't panicked

EDIT: `parsers.rs` isn't declared in lib.rs, intended?


Lonami:
> Wonder how CI doesn't panicked

This is an oversight. The client has the following features, but I forgot to add them to CI:

https://github.com/Lonami/grammers/blob/ea65eeecfecce3318f985df04fd09b40c4325ed6/lib/grammers-client/Cargo.toml#L16-L18

> `parsers.rs` isn't declared in lib.rs, intended?

...Might be another oversight from when I started to how the mtsender is used :)

PR #41, opened 2021-01-10
Builder changes

rupansh:
Use build() instead of impl Future
BannedRights kicks when Chat is group && view_messages is disallowed
AdminRights can only handle channels

### DIFF
--- a/lib/grammers-client/src/types/chats.rs
+++ b/lib/grammers-client/src/types/chats.rs
@@ -21,18 +21,14 @@ enum RightsChat {
     Unsupported,
 }
 
-impl From<Chat> for RightsChat {
-    fn from(chat: Chat) -> Self {
-        match chat {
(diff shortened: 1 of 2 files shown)


Lonami (changes requested):

lib/grammers-client/src/types/chats.rs
```diff
@@ -69,30 +63,74 @@ impl AdminRightsBuilder {
...
+                    self.rights = a.admin_rights.into();
+                    self.rank = a.rank.unwrap_or_else(String::new);
+                }
+                _ => (),
             }
-            tl::enums::ChannelParticipant::Admin(a) => {
-                self.rights = a.admin_rights.into();
-                self.rank = a.rank.unwrap_or_else(String::new);
+        } else if let RightsChat::Group(id) = &self.chat {
+            // TODO: Use iter_participants
```
Lonami, line 92: Is this ready to merge or do you want to fix this before?

PR #42, opened 2021-01-11
Some #[derive(Debug)] please otherwise i'm not going to understand anything

Kezii:
Was there a particular reason they were not there?


Lonami:
I think I didn't add it because the `Client` eventually had something that couldn't be `Debug`? Not sure. We probably want to review the derives on *all* types. For example, a lot of them make sense that they would also have `Clone`, some make sense as `PartialEq + Eq` too, maybe some `Ord`.

So I'm hesitant to merge this pull request, I would rather have one that double checks every single type and applies all derives where it makes sense, in a consistent way.

Lonami:
Closing, the derive would expose far more things than I'd like it to.

PR #47, opened 2021-02-05
Support another api for accessing photos in the message and downloading photo thumbs

AlexPl292:
Here is a draft pull request for the first review:
- Accessing photos via `message.photo()`
- Downloading thumbs via `message.photo().thumbs()[n].download(path)`
- Additional helper method for selecting a thumb: `message.photo().thumbs().largest().download(path)`

Things to finish:
- Docs
- Implement other thumbs downloading and decompression.


Lonami:
Sounds good, I'll review once you ping me for it.

AlexPl292:
Well, this should be ready for the first review :)

AlexPl292:
BTW, is there a way to write test for that?

lib/grammers-client/src/types/media.rs
```diff
@@ -26,9 +33,48 @@ pub struct Uploaded {
...
+    }
+
+    pub async fn download(&mut self, path: &Path) {
+        match &self.document.document {
+            Some(_) => {
+                if let Some(location) = self.to_input_location() {
+                    self.client
+                        .download_media_at_location(location, path)
+                        .await
+                        .unwrap();
```
Lonami, line 55:
```suggestion
                        .await?;
```

lib/grammers-client/src/types/media.rs
```diff
@@ -26,9 +33,48 @@ pub struct Uploaded {
...
 
+impl Document {
+    pub(crate) fn from_media(
+        document: tl::types::MessageMediaDocument,
+        client: ClientHandle,
+    ) -> Self {
+        Self { document, client }
+    }
+
+    pub async fn download(&mut self, path: &Path) {
```
Lonami, line 48: 1. This should return a `Result`, not panic on error.
2. We should do the same as `download_media_at_location` -- there's no reason not to.

```suggestion
    pub async fn download<P: AsRef<Path>>(&mut self, path: P) -> io::Result<()> {
```

lib/grammers-client/src/types/media.rs
```diff
@@ -26,9 +33,48 @@ pub struct Uploaded {
...
+        match &self.document.document {
+            Some(_) => {
+                if let Some(location) = self.to_input_location() {
+                    self.client
+                        .download_media_at_location(location, path)
+                        .await
+                        .unwrap();
+                }
+            }
+            None => (),
```
Lonami, line 58: We want to return `Err` here.

lib/grammers-client/src/types/media.rs
```diff
@@ -26,9 +33,48 @@ pub struct Uploaded {
...
+
+    pub async fn download(&mut self, path: &Path) {
+        match &self.document.document {
+            Some(_) => {
+                if let Some(location) = self.to_input_location() {
+                    self.client
+                        .download_media_at_location(location, path)
+                        .await
+                        .unwrap();
+                }
```
Lonami, line 56: Needs an `else` branch to return `Err`, as if it was `None`.

lib/grammers-client/src/types/message.rs
```diff
@@ -435,4 +436,12 @@ impl Message {
 
         None
     }
+
```
Lonami, line 439:
```suggestion

    /// Get document attached to the message if any.
    /// Note that this includes photos sent as files, videos, stickers, and anything sent as a file.
```

lib/grammers-client/src/types/message.rs
```diff
@@ -435,4 +436,12 @@ impl Message {
...
+    pub fn document(&self) -> Option<Document> {
+        if let Media::Document(document) = self.media()? {
+            return Some(document);
+        }
+
+        None
```
Lonami, line 445:
```suggestion
            Some(document)
        } else {
            None
        }
```

lib/grammers-client/src/types/media.rs
Lonami, line 48: Missing documentation here, please indicate that this is nothing but a shortcut for `ClientHandle::download_media_at_location`, along with the appropriated intra-doc links.

Lonami (changes requested):
Implementation looks mostly okay except for the panic and lack of documentation.

Lonami:
> BTW, is there a way to write tests for that?

For the time being, you can use doctests by writing example code much like it's done in the other friendly methods. It won't test that it runs, but it will test that it compiles, which is more than enough.

lib/grammers-client/src/types/media.rs
```diff
@@ -59,6 +59,24 @@ impl Document {
...
+                .iter()
+                .filter_map(|x| match x {
+                    tl::enums::DocumentAttribute::Filename(attr) => match attr {
+                        tl::types::DocumentAttributeFilename { file_name } => {
+                            Some(file_name.to_string())
+                        }
+                    },
+                    _ => None,
+                })
+                .nth(0),
```
Lonami, line 76:
```suggestion
                .next(),
```

Lonami (changes requested):

Lonami:
Make sure to add documentation to all `pub fn`. Accessors don't need code examples, but methods like `download` do.

AlexPl292:
Sorry for spamming, I've started to push some non-PR related commits, I forgot that they automatically attaching to this PR XD

I'll clean things up and ping you again.

Lonami:
> I forgot that they automatically attaching to this PR

As I understand it, the "correct" way to do this is to create a separate branch to work on a specific pull request, and then you can get back to master. But it's fine, I normally squash PRs anyway (unless they're large. where I appreciate a proper commit log).

AlexPl292:
Okay, I've removed trash commits, squashed the correct ones and added a few changes in a separate commit (added doc and updated `download` method). 
As I see, most of your comments are about the wrong code, sorry for wasting your time :\

lib/grammers-client/src/types/media.rs
```diff
@@ -63,6 +69,37 @@ impl Photo {
             P::Photo(photo) => photo.id,
         }
     }
+
+    /// Get photo thumbs.
```
Lonami, line 73: This could use a bit more information. What does "thumbs" stand for? Can I somehow rely on the order of the returned photos? If I have a photo media, can this be empty? What can I use the thumbnails for when I already have the photo? Can I use all the thumbnails to download a specific size, or some won't work?

Lonami (changes requested):

lib/grammers-client/src/types/photo_sizes.rs
```diff
@@ -0,0 +1,284 @@
...
+use crate::ClientHandle;
+
+pub enum PhotoSize {
+    Empty(SizeEmpty),
+    Size(Size),
+    Cached(CachedSize),
+    Stripped(StrippedSize),
+    Progressive(ProgressiveSize),
+    Path(PathSize),
+}
```
Lonami, line 17: I don't know if it's desirable to expose this much information. I think a `struct PhotoSize` with `impl` for `cached() -> bool`, `dim()` for dimensions, `width()`, `is_vector()` would be a lot more useful. What do you think?

In addition, I believe Telegram borrowed the `photo_type` string format from [VK Photo sizes format](https://vk.com/dev/photo_sizes). Maybe the `enum Size` for `Small`, `Medium`, etc. would be better (accidentally lost this comment as a part of the review and had to type it again).

lib/grammers-client/src/types/media.rs
AlexPl292: I've added a bit more docs

lib/grammers-client/src/types/photo_sizes.rs
AlexPl292: Well, my position is that good abstraction over the types should not hide the raw information. I mean, get more helper helpers (dim, width, is_vector) is great, but it's not okay if the user won't be able to get `StrippedSize` type. For example, maybe the user doesn't want to build a full featured application, but only a "super stripped photo sizes extractor".

But maybe this crate has a different purpose. If this is not about "you can build any type of application", but "you have a nice API to build applications fast and simple and for more complicated cases, please use raw calls and types", then we can provide only such methods.

Lonami: > If this is not about "you can build any type of application", but "you have a nice API to build applications fast and simple and for more complicated cases, please use raw calls and types", then we can provide only such methods.

Yes, that's it. The Rust community values semantic versioning a lot, and with Telegram changing things so often with new layers, most of the time it would be a breaking change. It would be silly to bump the major version of grammers on every new layer, because it technically is a breaking change. We'd end up with huge versions in no time, so semver becomes pointless.

`grammers-client` is a high-level client which you're not forced to use; you could rely on just `grammers-mtsender` and implement your abstractions your own way. And there will be a way to provide access to the raw types. But the public API should be simple, and avoid introducing breaking changes as much as possible.

Note I'm not saying we should hide `StrippedSize` away, we should offer a convenient way to access this information. `is_cached` would do well, as that's the main purpose of the type.

Lonami (approved):
Overall I think this is a good start. I'll merge it as is for now but you can improve upon it further if you wish.

lib/grammers-client/src/types/photo_sizes.rs
AlexPl292: Oh, thank you for the explanation. I didn't know about the issue with a frequent telegram schema updates.

Lonami: You can view all the layers and their changes at https://diff.telethon.dev/. Generally, there's a few every year (and sometimes they change the layer without incrementing the layer number). Providing a friendlier abstractions will make the client more "rusty" and also more resistant to changes.

PR #51, opened 2021-02-20
Fix license year in pre-commit

expectocode:


Lonami:
According to https://softwareengineering.stackexchange.com/a/130479, it doesn't really matter.

expectocode:
every day's a school day

PR #52, opened 2021-02-22
Do not download the same file chunk all the time

AlexPl292:
Without this fix, we download the safe file chunk on each next.


Lonami (approved):

Lonami:
That's quite a big mistake.

PR #54, opened 2021-03-16
Added the ability to set a data center 

Norbytus:


Lonami:
Thanks for the PR, but I have to ask, what's the rationale behind this change?

Norbytus:
> Thanks for the PR, but I have to ask, what's the rationale behind this change?

In the current implementation, the ip is sewn into an array, and it is not possible to switch to a specific one. I already wrote about this in the telegram chat, that it would be nice to be able to use a specific IP data center. For me, the main case is the use of a test data center.
image

Lonami:
I agree that there should be a way to provide a different IP address in order to tell the client to use the test servers, and I agree that hardcoding the test servers and saying "use `test: true` to go to the test servers" would be a bad idea because those IPs could change. However, I don't agree with the current implementation of this PR.

I don't think it's necessary to accept an entire array of addresses when only one is selected. `0` is also kind of special because it's the default, which would be awkward.

I think the PR should simply add `server_addr: Option<std::net::SocketAddr>` and use that to prevent the default selection otherwise. Not an IP-port pair, not a list, and not a DC ID.

The DC ID is kind of useless because, on login, the server will redirect you to your correct DC anyway, and then that's stored in the session file for later use.

Norbytus:
> I agree that there should be a way to provide a different IP address in order to tell the client to use the test servers, and I agree that hardcoding the test servers and saying "use `test: true` to go to the test servers" would be a bad idea because those IPs could change. However, I don't agree with the current implementation of this PR.
> 
> I don't think it's necessary to accept an entire array of addresses when only one is selected. `0` is also kind of special because it's the default, which would be awkward.
> 
> I think the PR should simply add `server_addr: Option<std::net::SocketAddr>` and use that to prevent the default selection otherwise. Not an IP-port pair, not a list, and not a DC ID.
> 
> The DC ID is kind of useless because, on login, the server will redirect you to your correct DC anyway, and then that's stored in the session file for later use.

Ok, i rewrite this

Norbytus:
In session i replace pair (IpV4addr, u16) to Into<SocketAddrs> maybe better use ToSocketAddrs but i am not sure

lib/grammers-client/src/client/client.rs
```diff
@@ -52,6 +53,7 @@ pub struct InitParams {
     // TODO catch up doesn't occur until we get an update that tells us if there was a gap, but
     // maybe we should forcibly try to get difference even if we didn't miss anything?
     pub catch_up: bool,
+    pub server_addr: Option<SocketAddr>,
```
Lonami, line 56: Could you add a doccomment on top of this field to explain what it does and why anyone would want to use it?

Lonami (changes requested):

lib/grammers-client/src/client/client.rs
Norbytus: Yep

lib/grammers-client/src/client/client.rs
```diff
@@ -52,6 +53,9 @@ pub struct InitParams {
     // TODO catch up doesn't occur until we get an update that tells us if there was a gap, but
     // maybe we should forcibly try to get difference even if we didn't miss anything?
     pub catch_up: bool,
+    ///Addr to telegram dc, by default use one of DC_ADDRESSES, but if you need a specific or test
+    ///you can use this param
```
Lonami, line 57:
```suggestion
    /// Server address to connect to. By default, the library will connect to the address stored
    /// in the session file (or a default production address if no such address exists). This
    /// field can be used to override said address, and is most commonly used to connect to one
    /// of Telegram's test servers instead.
```

lib/grammers-session/src/lib.rs
```diff
@@ -104,12 +104,27 @@ impl Session for MemorySession {
...
+                ipv4: if let Some(ip_v4) = ip_v4 {
+                    Some(i32::from_le_bytes(ip_v4.ip().octets()))
+                } else {
+                    None
+                },
+                ipv6: if let Some(ip_v6) = ip_v6 {
+                    Some(ip_v6.ip().octets())
+                } else {
+                    None
+                },
```
Lonami, line 126:
```suggestion
                ipv4: ip_v4.map(|addr| i32::from_le_bytes(addr.ip().octets())),
                ipv6: ip_v6.map(|addr| addr.ip().octets()),
```

PR #55, opened 2021-03-22
Add unpack_chat

the-blank-x:


Lonami:
While this is something useful to have right now, it isn't what I was aiming format. My idea of "packed chat" was "an object you can format as hex string or bytes", but not "something you can load an user from".

In my opinion, everything should work over `Chat` and not `PackedChat`. My idea was that you could get a `Chat` from a `PackedChat` and then use something like `chat.reload().await` to re-fetch the current state of the particular chat instance. This would be useful beyond just getting the username from a packed chat.

But maybe we could do both. What do you think?

the-blank-x:
You want to do something like `let chat = Chat::from_packed_chat(&packed_chat, handle.clone())` to have an instance of `Chat` with just only the ID and access hash, then do `chat.reload().await` to load all the information?

Lonami:
Yes, that's what I had envisioned. But then again it might be useful to provide this method as well. Telethon uses `get_entity`, in grammers I wanted to push away from the generic "entity" terminology and IMO "peer" is not as clear as "chat". Given we have `chat.pack()`, perhaps `client.unpack_chat` is the best name…

the-blank-x:
Do we implement `Chat::from_packed_chat` and `chat.reload` in this PR?

Lonami:
I'll merge this as-is. It's a good addition except for name bikeshedding which I can't be bothered to do at the moment.

PR #56, opened 2021-03-26
Photo may not be presented if the photo is already expired

AlexPl292:
Hi! It's not safe to unwrap a photo because it may be missing in cases when this is an expired photo. 

image

### DIFF
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -548,7 +548,7 @@ impl ClientHandle {
     /// let mut photos = client.iter_profile_photos(&chat);
     ///
     /// while let Some(photo) = photos.next().await? {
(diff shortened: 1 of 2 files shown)


Lonami:
The thing that annoys me is that Telegram has:
* Optional media.
* If there's photo media, the photo itself is optional.
* If the photo is present, the photo may be empty.
* Finally, you may actually have a photo.

I understand that there should be a way to check "media which has expired", but I don't think making the ID optional is the way to go. We also need a better way to handle "photo empty" (and any media, really). So I think I'll close this for now.

(If it's urgent because you encounter it a lot we could temporarily merge it though; let me know).

AlexPl292:
+1 about the structure.
No, it's fine to find an another way, I've just spent some time trying to understand what's wrong, so I did a PR with this fix.

PR #57, opened 2021-03-31
client: media: add sticker

rupansh:


lib/grammers-client/src/types/media.rs
```diff
@@ -157,6 +163,55 @@ impl Document {
...
+
+
+    /// If the document is a sticker,
+    /// Get attributes associated with the sticker
+    pub fn sticker_attributes(&self) -> Option<StickerAttributes> {
+        StickerAttributes::from_document(self)
+    }
+
+    /// Get the Mime Type of the document, if any
+    pub fn mime_type(&self) -> Option<&str> {
```
Lonami, line 175: I don't think this belongs to this PR.

lib/grammers-client/src/types/media.rs
```diff
@@ -157,6 +163,55 @@ impl Document {
                 .unwrap_or(""),
         }
     }
+
+
+    /// If the document is a sticker,
+    /// Get attributes associated with the sticker
+    pub fn sticker_attributes(&self) -> Option<StickerAttributes> {
```
Lonami, line 170: Is this the way to go? Why was the idea of having `Media::Sticker` discarded?

lib/grammers-client/src/types/media.rs
```diff
@@ -159,6 +167,40 @@ impl Document {
...
+                    document: document.clone(),
+                    attrs: sticker_attrs?,
+                    animated,
+                })
+            }
+            _ => None,
+        }
+    }
+
+    /// Get the first emoji associated with the sticker
```
Lonami, line 193:
```suggestion
    /// Get the emoji associated with the sticker.
```

PR #59, opened 2021-04-01
Add helper to get the smallest photo thumb

melix99:
My usecase: show a low resolution photo preview while the high resolution one is downloading.


Lonami:
Thanks for the PR, but I think I'd like to change the design of this a bit. Like probably an `enum` for the size so it just picks the most appropriated size (imagine `photo.download(PhotoSize::Smallest)`, I think).

PR #61, opened 2021-04-06
InputMessage: add option to use media from an existing message

rupansh:


Lonami (approved):
LGTM, but the user can't change the `ttl_seconds` at the moment.

PR #62, opened 2021-04-06
Markdown line breaks

Marekkon5:
Parse line breaks in markdown correctly.


lib/grammers-client/src/parsers.rs
```diff
@@ -138,6 +138,21 @@ pub fn parse_markdown_message(message: &str) -> (String, Vec<tl::enums::MessageE
...
 
+        // \n
+        Event::SoftBreak => {
+            text.push('\n');
+            offset += 1;
+        }
+        // \\\n or   \n
+        Event::HardBreak => {
+            text.push_str("\n\n");
+            offset += 1;
```
Lonami, line 149:
```suggestion
            offset += 2;
```

PR #66, opened 2021-04-14
client: Introduce SessionGuard

rupansh:
implements Client::session


rupansh:
Will go with an alternative

PR #67, opened 2021-04-22
Add PackedChat::to_bytes and PackedChat::from_bytes

the-blank-x:


Lonami:
Just realized the downside to this is that we have no way of knowing if a `Channel` is a `megagroup` or a `broadcast` (and I guess `gigagroup` now too). It would also probably be useful to store if the user is a bot or not.

the-blank-x:
Are there perks of knowing a channel is a megagroup, broadcast or gigagroup, or that a user is a bot?

Lonami:
At the very least, we need megagroup/broadcast, because when constructing the `Chat` we differentiate between groups or broadcast channels.

Lonami:
Note that I've already fixed this locally, just didn't push yet.

PR #68, opened 2021-05-02
feat(client): add `attribute` method to `InputMessage`

aykxt:


lib/grammers-client/src/types/attributes.rs
```diff
@@ -0,0 +1,58 @@
...
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+use grammers_tl_types as tl;
+
+pub enum Attribute {
+    Audio {
+        voice: bool,
```
Lonami, line 12: I'd go as far as separating this into another variant (why'd you provide album for a voice recording?)

lib/grammers-client/src/types/attributes.rs
```diff
@@ -0,0 +1,58 @@
...
+    Video {
+        round_message: bool,
+        supports_streaming: bool,
+        duration: i32,
+        w: i32,
+        h: i32,
+    },
+}
+
+impl From<&Attribute> for tl::enums::DocumentAttribute {
```
Lonami, line 27: Perhaps we should consume the attribute instead of taking it by reference?

lib/grammers-client/src/types/input_message.rs
```diff
@@ -151,6 +153,19 @@ impl InputMessage {
         self
     }
 
+    pub fn attribute(mut self, attr: &Attribute) -> Self {
```
Lonami, line 156: This will need documentation.

lib/grammers-client/src/types/attributes.rs
aykxt: done

lib/grammers-client/src/types/attributes.rs
aykxt: I was thinking of that too, done

lib/grammers-client/src/types/attributes.rs
```diff
@@ -0,0 +1,70 @@
...
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+use grammers_tl_types as tl;
+
+pub enum Attribute {
+    Audio {
+        duration: i32,
```
Lonami, line 12: Would it make sense to provide a `Duration` here, even if we convert it to seconds later?

lib/grammers-client/src/types/attributes.rs
```diff
@@ -0,0 +1,70 @@
...
+        title: Option<String>,
+        performer: Option<String>,
+    },
+    Voice {
+        duration: i32,
+        waveform: Option<Vec<u8>>,
```
Lonami, line 18: This will need some explanation on how waveforms are to be calculated/rendered…

aykxt: Actually, I have no idea how they are calculated  :D

lib/grammers-client/src/types/attributes.rs
aykxt: `Duration` accepts u64 for seconds, while the API allows up to i32, which should not be exceeded in a real life scenario, but the type cast is still prone to errors 🤔

Lonami: I'm okay with a panic here. No audio should last that long.

lib/grammers-client/src/types/attributes.rs
Lonami: Unfortunately, I think the rendering is inconsistent, but at least on Telegram Desktop it's known.

lib/grammers-client/src/types/attributes.rs
aykxt: Okay, it will panic now at values greater than i32 :: MAX

Lonami:
Next up documentation about what attributes are, why you'd want to use them, and an example.

lib/grammers-client/src/types/input_message.rs
```diff
@@ -152,6 +151,29 @@ impl InputMessage {
         self
     }
 
+    /// Add additional attributes to the message.
```
Lonami, line 154:
```suggestion
    /// Add additional attributes to the message.
    ///
    /// This must be called *after* setting a file.
```

Lonami:
Thanks.

PR #69, opened 2021-05-10
Add support for contact attachments

AlexPl292:
It looks like that contact attachment is pretty easy to implement, but I'm not entirely sure if this is what you expect from this library. Let me know your thoughts on it.

Also, I don't know if you expect docs for `first_name, `last_name` and other getters for this case. For me, they look pretty obvious.

### DIFF
--- a/lib/grammers-client/src/types/media.rs
+++ b/lib/grammers-client/src/types/media.rs
@@ -36,11 +36,17 @@ pub struct Uploaded {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+pub struct Contact {
+    contact: tl::types::MessageMediaContact,
+}
+
+#[derive(Clone, Debug, PartialEq)]
 #[non_exhaustive]
 pub enum Media {
     Photo(Photo),
     Document(Document),
     Sticker(Sticker),
+    Contact(Contact),
 }
 
 impl Photo {
@@ -267,6 +273,37 @@ impl Sticker {
     }
 }
 
+impl Contact {
+    pub(crate) fn from_media(contact: tl::types::MessageMediaContact) -> Self {
+        Self { contact }
+    }
+
+    pub(crate) fn to_input_media(&self) -> tl::types::InputMediaContact {
+        tl::types::InputMediaContact {
+            phone_number: self.contact.phone_number.clone(),
+            first_name: self.contact.first_name.clone(),
+            last_name: self.contact.last_name.clone(),
+            vcard: self.contact.vcard.clone(),
+        }
+    }
+
+    pub fn phone_number(&self) -> String {
+        self.contact.phone_number.clone()
+    }
+
+    pub fn first_name(&self) -> String {
+        self.contact.first_name.clone()
+    }
+
+    pub fn last_name(&self) -> String {
+        self.contact.last_name.clone()
+    }
+
+    pub fn vcard(&self) -> String {
+        self.contact.vcard.clone()
+    }
+}
+
 impl Uploaded {
     pub(crate) fn from_raw(input_file: tl::enums::InputFile) -> Self {
         Self { input_file }
@@ -289,7 +326,7 @@ impl Media {
             M::Empty => None,
             M::Photo(photo) => Some(Self::Photo(Photo::from_media(photo, client))),
             M::Geo(_) => None,
-            M::Contact(_) => None,
+            M::Contact(contact) => Some(Self::Contact(Contact::from_media(contact))),
             M::Unsupported => None,
             M::Document(document) => {
                 let document = Document::from_media(document, client);
@@ -314,6 +351,7 @@ impl Media {
             Media::Photo(photo) => photo.to_input_media().into(),
             Media::Document(document) => document.to_input_media().into(),
             Media::Sticker(sticker) => sticker.document.to_input_media().into(),
+            Media::Contact(contact) => contact.to_input_media().into(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,37 @@ impl Sticker {
...
+            phone_number: self.contact.phone_number.clone(),
+            first_name: self.contact.first_name.clone(),
+            last_name: self.contact.last_name.clone(),
+            vcard: self.contact.vcard.clone(),
+        }
+    }
+
+    pub fn phone_number(&self) -> String {
+        self.contact.phone_number.clone()
+    }
```
aykxt, line 292: Perhaps we should return a string slice here, instead of cloning it. same below.
```suggestion
    pub fn phone_number(&self) -> &str {
        self.contact.phone_number.as_str()
    }
```

lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,37 @@ impl Sticker {
...
+        self.contact.first_name.clone()
+    }
+
+    pub fn last_name(&self) -> String {
+        self.contact.last_name.clone()
+    }
+
+    pub fn vcard(&self) -> String {
+        self.contact.vcard.clone()
+    }
```
aykxt, line 304:
```suggestion
    pub fn first_name(&self) -> &str {
        self.contact.first_name.as_str()
    }

    pub fn last_name(&self) -> &str {
        self.contact.last_name.as_str()
    }

    pub fn vcard(&self) -> &str {
        self.contact.vcard.as_str()
    }
```

lib/grammers-client/src/types/media.rs
Lonami: Yes, it's better if we return a `&str` here (and the other methods too; the user can clone it if they need the owned version).

lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,44 @@ impl Sticker {
...
+            last_name: self.contact.last_name.clone(),
+            vcard: self.contact.vcard.clone(),
+        }
+    }
+
+    pub fn phone_number(&self) -> &str {
+        self.contact.phone_number.as_str()
+    }
+
+    /// May be empty if it's not set by sender.
```
Lonami, line 294: Can this actually be empty?

Lonami:
```python
>>> cl.send_file(b, types.InputMediaContact(phone_number='', first_name='', last_name='', vcard=''))
telethon.errors.rpcerrorlist.MediaEmptyError: The provided media object is invalid or the current account may not be able to send it (such as games as users) (caused by SendMediaRequest)
>>> cl.send_file(b, types.InputMediaContact(phone_number='a', first_name='', last_name='', vcard=''))
telethon.errors.rpcerrorlist.MediaEmptyError: The provided media object is invalid or the current account may not be able to send it (such as games as users) (caused by SendMediaRequest)
>>> cl.send_file(b, types.InputMediaContact(phone_number='1', first_name='', last_name='', vcard=''))
# worked
>>> cl.send_file(b, types.InputMediaContact(phone_number='a1a', first_name='', last_name='', vcard=''))
# worked
```

Looks like Telegram removes non-digits from the phone number, and if empty, fails. But the first name can actually be empty.

lib/grammers-client/src/types/media.rs
Lonami, line 294:
```suggestion
    /// The contact's first name. Although official clients will always send a non-empty string,
    /// it is possible for this field to be empty when sent via different means.
```

lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,44 @@ impl Sticker {
...
+    pub fn first_name(&self) -> &str {
+        self.contact.first_name.as_str()
+    }
+
+    /// May be empty if it's not set by sender.
```
Lonami, line 299:
```suggestion
    /// The contact's last name. May be empty if it's not set by sender.
```

lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,44 @@ impl Sticker {
...
+    pub fn last_name(&self) -> &str {
+        self.contact.last_name.as_str()
+    }
+
+    /// vCard is a format standard for electronic business cards.
+    ///
+    /// See https://en.wikipedia.org/wiki/VCard
+    ///
+    /// May be empty if it's not set by sender.
```
Lonami, line 308:
```suggestion
    /// Contact information in [vCard format][1]. Applications such as Telegram Desktop leave this
    /// field empty. The vCard version used in this field could be any. The field may also contain
    /// arbitrary text when sent by non-official clients.
    ///
    /// [1]: https://en.wikipedia.org/wiki/VCard
```

lib/grammers-client/src/types/media.rs
```diff
@@ -267,6 +273,44 @@ impl Sticker {
...
+    pub(crate) fn to_input_media(&self) -> tl::types::InputMediaContact {
+        tl::types::InputMediaContact {
+            phone_number: self.contact.phone_number.clone(),
+            first_name: self.contact.first_name.clone(),
+            last_name: self.contact.last_name.clone(),
+            vcard: self.contact.vcard.clone(),
+        }
+    }
+
+    pub fn phone_number(&self) -> &str {
```
Lonami, line 290:
```suggestion
    /// The contact's phone number, in international format. This field will always be a non-empty
    /// string of digits, although there's no guarantee that the number actually exists.
    pub fn phone_number(&self) -> &str {
```

Lonami (changes requested):

Lonami (approved):

Lonami:
Thanks.

PR #70, opened 2021-05-23
Fix linting errors

aykxt:
This PR fixes some linting errors and removes unnecessary mutable borrows.

### DIFF
--- a/lib/grammers-client/src/utils.rs
+++ b/lib/grammers-client/src/utils.rs
@@ -27,7 +27,9 @@ pub(crate) fn generate_random_id() -> i64 {
             .expect("system time is before epoch")
             .as_nanos() as i64;
 
-        drop(LAST_ID.compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst));
+        LAST_ID
+            .compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst)
+            .unwrap();
     }
 
     LAST_ID.fetch_add(1, Ordering::SeqCst)
--- a/lib/grammers-mtsender/src/lib.rs
+++ b/lib/grammers-mtsender/src/lib.rs
@@ -470,7 +470,7 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
                         };
 
                         let req = self.requests.remove(i);
-                        req.result.send(result).unwrap();
+                        drop(req.result.send(result));
                         break;
                     }
                     _ => {}
--- a/lib/grammers-mtsender/src/lib.rs
+++ b/lib/grammers-mtsender/src/lib.rs
@@ -34,7 +34,7 @@ use tokio::time::{sleep_until, Duration, Instant};
 /// Telegram will close the connection with roughly a megabyte of data,
 /// so to account for the transports' own overhead, we add a few extra
 /// kilobytes to the maximum data size.
-const MAXIMUM_DATA: usize = (1 * 1024 * 1024) + (8 * 1024);
+const MAXIMUM_DATA: usize = (1024 * 1024) + (8 * 1024);
 
 /// Every how often are pings sent?
 const PING_DELAY: Duration = Duration::from_secs(60);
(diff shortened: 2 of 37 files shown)


lib/grammers-client/src/utils.rs
```diff
@@ -27,7 +27,9 @@ pub(crate) fn generate_random_id() -> i64 {
             .expect("system time is before epoch")
             .as_nanos() as i64;
 
-        drop(LAST_ID.compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst));
+        LAST_ID
+            .compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst)
+            .unwrap();
```
Lonami, line 32: This is semantically different, but I guess in this case it's always going to be `Ok`.

lib/grammers-mtsender/src/lib.rs
```diff
@@ -58,7 +58,9 @@ pub(crate) fn generate_random_id() -> i64 {
             .expect("system time is before epoch")
             .as_nanos() as i64;
 
-        drop(LAST_ID.compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst));
+        LAST_ID
+            .compare_exchange(0, now, Ordering::SeqCst, Ordering::SeqCst)
+            .unwrap();
```
Lonami, line 63: Also semantically different.

lib/grammers-mtsender/src/lib.rs
```diff
@@ -467,9 +469,8 @@ impl<T: Transport, M: Mtp> Sender<T, M> {
                             }
                         };
 
-                        drop(req);
                         let req = self.requests.remove(i);
-                        drop(req.result.send(result));
+                        req.result.send(result).unwrap();
```
Lonami, line 473: I don't think we can `unwrap` this one. The user may have dropped the oneshot receiver by the time a response from Telegram arrives. In this case we actually want to ignore. If clippy is complaining, perhaps we could `debug!()` `if .is_err()`.

lib/grammers-session/src/lib.rs
```diff
@@ -120,7 +120,7 @@ impl Session {
             .next()
     }
 
-    pub fn insert_dc(&self, id: i32, server_addr: SocketAddr, auth: [u8; 256]) {
+    pub fn insert_dc(&self, id: i32, addr: SocketAddr, auth: [u8; 256]) {
```
Lonami, line 123: Why?

Lonami (changes requested):
I would have preferred if the `mut` changes were a separate PR or, at least, a separate commit. Same for the semantic changes regarding `drop`.

That said, after addressing the `drop` changes, I'll merge it regardless if separating the `mut` stuff into another PR is too much effort.

lib/grammers-session/src/lib.rs
aykxt: I think it's obvious which address is meant, also server_addr was aliased to addr later in the code. Should I rename it back to server_addr?

lib/grammers-mtsender/src/lib.rs
aykxt: so leave it to another PR and revert or should I add the debug log?

aykxt:
> That said, after addressing the `drop` changes, I'll merge it regardless if separating the `mut` stuff into another PR is too much effort.

I think it's not worth separating the `mut` changes into a different PR (they are also some kind of linting errors, aren't they?).

lib/grammers-session/src/lib.rs
Lonami: Nah it's fine, I just wondered.

lib/grammers-mtsender/src/lib.rs
Lonami: Please remove this change from this one and then we can send a separate one with a better "fix".

Lonami:
> [...] the `mut` changes [...] (they are also some kind of linting errors, aren't they?).

No, they're also semantic changes, not just pure stylistic changes from the clippy lints. However, if it's too much effort to separate them, I can still merge as is.

lib/grammers-mtsender/src/lib.rs
aykxt: done

Lonami:
Thanks.

PR #72, opened 2021-05-24
client: chat builders: impl Future

rupansh:
World's first Zero Cost™ Async Builder

### DIFF
--- a/lib/grammers-client/DEPS.md
+++ b/lib/grammers-client/DEPS.md
@@ -73,4 +73,5 @@ Used to test that this file lists all dependencies from `Cargo.toml`.
 
 ## pin-project-lite
 
-Used for Zero Cost Async Builders
+Used for return custom types that `impl Future` so that the requests can be further configured
+without having to use `Box`.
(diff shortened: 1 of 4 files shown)


lib/grammers-client/DEPS.md
```diff
@@ -70,3 +70,7 @@ Used by the examples to showcase how one may configure logging for more informat
 ## toml
 
 Used to test that this file lists all dependencies from `Cargo.toml`.
+
+## pin-project-lite
+
+Used for Zero Cost Async Builders
```
Lonami, line 76:
```suggestion
Used for return custom types that `impl Future` so that the requests can be further configured
without having to use `Box`.
```

lib/grammers-client/src/types/chats.rs
```diff
@@ -27,54 +31,134 @@ pub struct AdminRightsBuilder {
...
+                    other: false,
+                },
+            }),
+            gen,
+            fut: None,
         }
     }
 
+    fn inner_mut<'a>(&'a mut self) -> &'a mut AdminRightsBuilderInner {
+        // Unwrap safety: AdminRightsBuilderInner should never be None unless polled after being
```
Lonami, line 132: Does the "safety" here mean it would be `unsafe` to `panic!`?

rupansh: > Does the "safety" here mean it would be `unsafe` to `panic!`?

No it means inner_mut will never panic.
`.await` will consume the builder. (i.e self.inner.poll is called)
`self.inner` only becomes `None` after being `poll`ed.

PR #73, opened 2021-05-25
Improve upload speed performance

aykxt:
With this PR, the upload speed is about three times faster. Actually this could be further improved by creating new connections, but I'll leave that to another PR.

### DIFF
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -9,16 +9,27 @@
 use crate::types::{Media, Uploaded};
 use crate::utils::generate_random_id;
 use crate::Client;
+use futures_util::future::try_join_all;
 use grammers_mtsender::InvocationError;
 use grammers_tl_types as tl;
-use std::io::SeekFrom;
-use std::path::Path;
-use tokio::fs;
-use tokio::io::{self, AsyncRead, AsyncReadExt as _, AsyncSeekExt as _, AsyncWriteExt as _};
+use std::{
+    io::SeekFrom,
+    path::Path,
+    sync::{
+        atomic::{AtomicI32, Ordering},
+        Arc,
+    },
+};
+use tokio::{
+    fs,
+    io::{self, AsyncRead, AsyncReadExt as _, AsyncSeekExt as _, AsyncWriteExt as _},
+    sync::Mutex,
+};
 
 pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
 const BIG_FILE_SIZE: usize = 10 * 1024 * 1024;
+const WORKER_COUNT: usize = 4;
 
 pub struct DownloadIter {
     client: Client,
@@ -209,9 +220,9 @@ impl Client {
     /// ```
     ///
     /// [`InputMessage`]: crate::types::InputMessage
-    pub async fn upload_stream<S: AsyncRead + Unpin>(
+    pub async fn upload_stream<S: AsyncRead + Unpin + Send + 'static>(
         &self,
-        stream: &mut S,
+        stream: S,
         size: usize,
         name: String,
     ) -> Result<Uploaded, io::Error> {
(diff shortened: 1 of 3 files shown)


JuniorJPDJ:
I'm interested in some numbers. Do you have something to provide? :D

aykxt:
> I'm interested in some numbers. Do you have something to provide? :D

Tested on a 1Gbit/s internet connection (23.2 MB file):
before: ~9000ms, after: ~2500-3000ms

JuniorJPDJ:
10MBps? Seems very close to Telegram Desktop scores! It's much better than telethon's! :DDD

Lonami:
> It's much better than telethon's! :DDD

Rude.

lib/grammers-client/src/client/files.rs
```diff
@@ -9,16 +9,27 @@
...
+    sync::{
+        atomic::{AtomicI32, Ordering},
+        Arc,
+    },
+};
+use tokio::{
+    fs,
+    io::{self, AsyncRead, AsyncReadExt as _, AsyncSeekExt as _, AsyncWriteExt as _},
+    sync::Mutex,
+};
```
Lonami, line 27: The codebase is not using this import style anywhere else, so I think we should change it to be consistent to the non-nested form.

lib/grammers-client/src/client/files.rs
```diff
@@ -9,16 +9,27 @@
...
 
 pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
 const BIG_FILE_SIZE: usize = 10 * 1024 * 1024;
+const WORKER_COUNT: usize = 4;
```
Lonami, line 32: This should be configurable. With the work done by #72, it could maybe be configurable (although we can change that later).

lib/grammers-client/src/client/files.rs
```diff
@@ -325,6 +343,60 @@ impl Client {
...
+            current_part: AtomicI32::new(0),
+        }
+    }
+
+    fn total_parts(&self) -> i32 {
+        self.total_parts
+    }
+
+    async fn next_part(&self) -> Result<Option<(i32, Vec<u8>)>, io::Error> {
+        let mut stream = self.stream.lock().await;
```
Lonami, line 371: Maybe we should use our custom mutex here so that they can be debugged too.

lib/grammers-client/src/client/files.rs
```diff
@@ -223,71 +234,78 @@ impl Client {
...
-                        ));
+        let parts = Arc::new(PartStream::new(stream, size));
+        let total_parts = parts.total_parts();
+
+        if big_file {
+            let mut tasks = Vec::with_capacity(WORKER_COUNT);
+            for _ in 0..WORKER_COUNT {
+                let handle = self.clone();
+                let parts = Arc::clone(&parts);
+                let task = tokio::task::spawn(async move {
```
Lonami, line 245: This `spawn` is somewhat unfortunate, as nowhere else in the code tasks are spawned. Would it be possible to do without? I'm trying to avoid it because it's executor-specific, and I'd like to move to a `futures`-only approach at some point (or the best I could).

lib/grammers-client/src/client/files.rs
```diff
@@ -325,6 +343,60 @@ impl Client {
...
+        }
+    }
+
+    fn total_parts(&self) -> i32 {
+        self.total_parts
+    }
+
+    async fn next_part(&self) -> Result<Option<(i32, Vec<u8>)>, io::Error> {
+        let mut stream = self.stream.lock().await;
+        let current_part = self.current_part.fetch_add(1, Ordering::SeqCst);
```
Lonami, line 372: Is the atomic needed if we're locking the `stream`?

lib/grammers-client/src/client/files.rs
aykxt: Hmm, how do you want to do that without spawning tasks?

lib/grammers-client/src/client/files.rs
aykxt: Removed it.

lib/grammers-client/src/client/files.rs
aykxt: Done

lib/grammers-client/src/client/files.rs
aykxt: Done.

lib/grammers-client/src/client/files.rs
```diff
@@ -325,6 +335,70 @@ impl Client {
...
+        }
+    }
+
+    fn total_parts(&self) -> i32 {
+        self.total_parts
+    }
+
+    async fn next_part(&self) -> Result<Option<(i32, Vec<u8>)>, io::Error> {
+        let mut lock = self.inner.lock("read part").await;
+        if lock.current_part >= self.total_parts {
```
Lonami, line 373: Not what I had in mind w.r.t. getting rid of the atomic but fair enough. I presume this is a bit safer than having the parts without a lock.

Lonami (approved):

lib/grammers-client/src/client/files.rs
aykxt: Since the part count is always used when the lock is held, there should be no problem here. Also mutating directly the part count is unsafe and therefore must be behind some kind of mutex.

PR #74, opened 2021-05-28
Add a friendly `Update::InlineQuery`

aykxt:


lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,169 @@
...
+impl InlineQuery {
+    pub(crate) fn new(client: &Client, query: tl::types::UpdateBotInlineQuery) -> Self {
+        Self {
+            query,
+            client: client.clone(),
+        }
+    }
+
+    // User that sent the query.
+    pub fn user_id(&self) -> i32 {
```
Lonami, line 27: Can't this return the entire `Chat`? Is the user not in the entity map?

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,169 @@
...
+            },
+            client: self.client.clone(),
+        }
+    }
+}
+
+impl Answer {
+    /// Whether the results should show as a gallery (grid) or not.
+    /// Defaults to false.
+    pub fn gallery(mut self, gallery: bool) -> Self {
```
Lonami, line 57: Do we want a `bool` input parameter on this one, or would it be enough to have it parameter-less and calling it sets it to `true`?

I guess it depends. How often will `gallery` be determined programmatically? If it's not a lot, `.gallery()` may be more comfortable to write than `.gallery(true)`. (The same applies to `private`.)

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // The text of the inline query.
+    pub fn text(&self) -> &str {
+        self.query.query.as_str()
+    }
+
+    /// Answer the inline query.
+    // TODO: add example
+    pub fn answer(&self, results: Vec<tl::enums::InputBotInlineResult>) -> Answer {
```
Lonami, line 38: I'd prefer having a `struct InlineResult(tl::enums::InputBotInlineResult)`, and then the custom types like `Article` implement into `InlineResult`. I would like to avoid having any raw type (from `tl::`) in the public API at all.

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,169 @@
...
+            tl::types::InlineBotSwitchPm {
+                text: text.into(),
+                start_param: start_param.into(),
+            },
+        ));
+        self
+    }
+
+    /// Answers the inline query with the given results.
+    pub async fn send(self) -> Result<(), InvocationError> {
```
Lonami, line 97: I'd like to be able to `await` on the `Answer` directly, much like #72, but I guess we can do that after merging this with `send`.

lib/grammers-client/src/types/inline_query.rs
aykxt: I don't now if inline queries are bind to a specific chat, however I could only find `user_id` [here](https://core.telegram.org/constructor/updateBotInlineQuery).

lib/grammers-client/src/types/inline_query.rs
aykxt: I've been thinking about that too, perhaps calling `.gallery()` is the better way to go. (I really wish Rust supported optional arguments, it would make things a lot easier!)

lib/grammers-client/src/types/inline_query.rs
Lonami: Individual updates themselves don't include the chats, because a single batch may share the same chat for many updates, so it would be wasteful. Instead, the `Updates` constructor is the one with the chats, which for the time being, is an `Arc<ChatMap>`:

https://github.com/Lonami/grammers/blob/f2106e8e234b021819b42e2244d111c72c6279d0/lib/grammers-client/src/types/callback_query.rs#L36-L46

We probably can do something similar here (Telegram *should* send the user). Regardless, we should always aim to return a `Chat`, for consistency, even if it really can only have the `id` field.

aykxt: Done.

lib/grammers-client/src/types/inline_query.rs
aykxt: Done.

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// An inline query answer builder.
+pub struct Answer {
+    request: tl::functions::messages::SetInlineBotResults,
+    client: Client,
+}
+
+/// An inline query result.
+///
+/// grammers currently supports results of the following types that implement [`Into<InlineResult>`]:
```
Lonami, line 23: I don't recall other doc-comments mentioning the name of the library, so it sounds a bit weird.

```suggestion
/// The following types that implement [`Into<InlineResult>`]:
```

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,196 @@
...
+    // User that sent the query.
+    pub fn chat(&self) -> &Chat {
+        self.chats
+            .get(
+                &tl::types::PeerUser {
+                    user_id: self.query.user_id,
+                }
+                .into(),
+            )
+            .unwrap()
```
Lonami, line 55: Damn `rustfmt` can produce some ugly stuff.

lib/grammers-client/src/types/inline_query.rs
```diff
@@ -0,0 +1,196 @@
...
+    ) -> Self {
+        Self {
+            query,
+            client: client.clone(),
+            chats: chats.clone(),
+        }
+    }
+
+    // User that sent the query.
+    pub fn chat(&self) -> &Chat {
```
Lonami, line 47: I do wonder if we should return the friendly `User` from methods like these. `chat` is definitely confusing though, because it's not the chat where the inline query was executed. `sender` is consistent with `Message` and now also `CallbackQuery`.

```suggestion
    pub fn sender(&self) -> &Chat {
```

aykxt: It would probably make more sense, since the sender is always a user.

PR #76, opened 2021-05-30
Support external documents and photos

aykxt:


lib/grammers-client/src/types/input_message.rs
```diff
@@ -134,6 +134,23 @@ impl InputMessage {
         self
     }
 
+    /// Include an external photo in the message.
+    ///
+    /// The server will download and compress the image and convert it to JPEG format if
+    /// necessary.
```
Lonami, line 140:
```suggestion
    /// The Telegram server will download and compress the image and convert it to JPEG format if necessary.
```

lib/grammers-client/src/types/input_message.rs
```diff
@@ -158,6 +175,22 @@ impl InputMessage {
         self
     }
 
+    /// Include an external file as a document in the message.
+    ///
+    /// You can use this to send videos, stickers, audios, or uncompressed photos.
+    ///
```
Lonami, line 181:
```suggestion
    ///
    /// The Telegram server will download and compress the image and convert it to JPEG format if necessary.
    ///
    ///
```

lib/grammers-client/src/types/input_message.rs
aykxt: I copied the text from the `.photo()` method. Should I add the "Telegram" there too?

Lonami: Oh, didn't even realize. Sure.

PR #78, opened 2021-05-31
Further improve upload performance

aykxt:
This PR continues the work done in #73 and further improves the upload performance drastically by creating multiple connections to the telegram servers. (Upload of a 23MB takes now ~1100ms (before ~2700-3000ms) on a 1Gbit/s connection)

### DIFF
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -22,6 +22,7 @@ pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
 const BIG_FILE_SIZE: usize = 10 * 1024 * 1024;
 const WORKER_COUNT: usize = 4;
+const CONNECTION_COUNT: usize = 3;
 
 pub struct DownloadIter {
     client: Client,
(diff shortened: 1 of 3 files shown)


JuniorJPDJ:
wow. WOW :o

lib/grammers-client/src/client/files.rs
```diff
@@ -231,35 +232,44 @@ impl Client {
...
-
-                        if !ok {
-                            return Err(io::Error::new(
-                                io::ErrorKind::Other,
-                                "server failed to store uploaded data",
-                            ));
-                        }
+                let connection = async move {
+                    let mut tasks = Vec::with_capacity(WORKER_COUNT);
+                    let client = Client::connect(self.0.config.clone()).await.unwrap();
```
Lonami, line 240: I need to look deeper into this, but I'm not sure creating an entire new client is the route we want to go. (The `unwrap()` definitely needs to go though.)

This also introduces an ever greater chance for things to go wrong (what if one connection fails but the others don't?).

lib/grammers-client/src/client/files.rs
```diff
@@ -231,35 +232,44 @@ impl Client {
...
+                                if !ok {
+                                    return Err(io::Error::new(
+                                        io::ErrorKind::Other,
+                                        "server failed to store uploaded data",
+                                    ));
+                                }
+                            }
+                            Ok(())
+                        };
+                        tasks.push(task);
```
Lonami, line 265: I think we should call https://core.telegram.org/mtproto/service_messages#request-to-destroy-session before dropping the extra clients?

When creating a client, deeper down, a random session ID is generated and used. We may just need to add a method on the `mtsender` to invoke that.

lib/grammers-client/src/client/files.rs
```diff
@@ -231,35 +232,44 @@ impl Client {
...
+                let connection = async move {
+                    let mut tasks = Vec::with_capacity(WORKER_COUNT);
+                    let client = Client::connect(self.0.config.clone()).await.unwrap();
+                    for _ in 0..WORKER_COUNT {
+                        let handle = client.clone();
+                        let parts = Arc::clone(&parts);
+                        let task = async move {
+                            while let Some((part, bytes)) = parts.next_part().await? {
+                                let ok = handle
+                                    .invoke(&tl::functions::upload::SaveBigFilePart {
```
Lonami, line 247: Yeah we're lucky this doesn't trigger updates through the corresponding socket (https://core.telegram.org/api/updates#subscribing-to-updates).

lib/grammers-client/src/client/files.rs
aykxt: We could return an IO error instead, the try join method will fail immediately if any of the clients failed to connect.

Lonami: Would it not be better to fallback to a single connection approach? Or however many workers managed to connect.

Also, this is not a pool. If my bot receives a command from a hundred users at once and I need to serve them all at the same time, this would create 300 connections.

JuniorJPDJ:
[*]

PR #82, opened 2021-07-16
(#81) Fix Deadlock in client.next_update

### LINKED ISSUE #81: Deadlock in client.next_update due to double locking chat_hashes
The [client.next_update](https://github.com/Lonami/grammers/blob/146d2bc965b91629badce4f537fd1cb95ccf6ff1/lib/grammers-client/src/client/updates.rs#L48-L100) function locks the `chat_hashes` resource twice: initially to invoke `message_box.get_channel_difference` and then to invoke `message_box.apply_channel_difference`.

The first MutexGuard [1] is not dropped before trying to acquire the resource the second time [2] therefore causing a deadlock.

[1]
https://github.com/Lonami/grammers/blob/146d2bc965b91629badce4f537fd1cb95ccf6ff1/lib/grammers-client/src/client/updates.rs#L68

[2]
https://github.com/Lonami/grammers/blob/146d2bc965b91629badce4f537fd1cb95ccf6ff1/lib/grammers-client/src/client/updates.rs#L77-L80

novelhawk:
Fixes #81

### DIFF
--- a/lib/grammers-client/src/client/updates.rs
+++ b/lib/grammers-client/src/client/updates.rs
@@ -64,20 +64,14 @@ impl Client {
                 continue;
             }
 
-            let chat_hashes = self.0.chat_hashes.lock("client.next_update");
+            let mut chat_hashes = self.0.chat_hashes.lock("client.next_update/get_channel_difference");
             if let Some(request) = message_box.get_channel_difference(&chat_hashes) {
                 drop(message_box);
                 let response = self.invoke(&request).await?;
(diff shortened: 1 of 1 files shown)


novelhawk:
Unless it's needed for weird reason the double lock could be merged into a single lock. Waiting for comment.

novelhawk:
After discussing on the official Telegram channel, we've determined that there is no need for the `next_update` Future to be `Send` therefore we can merge the two locks in a single one.

PR #85, opened 2021-08-20
Faster decryption

Marekkon5:
Hello, I've replaced some of the Rust iterators with plain loops in `ige_decrypt`, so the compiler can optimize it more and use SIMD. I've tried patching `ige_encrypt`, however no matter what I tried I always got worse results. I was once able to get better one, but wasn't able to reproduce it again, so am leaving the original function in.

Benchmarks:
The improvement is ~250MB/s on bigger chunks.

left = grammers master, right = my fork

Thanks.


lib/grammers-crypto/src/aes.rs
```diff
@@ -12,7 +12,6 @@ use aes::cipher::{BlockDecrypt, BlockEncrypt, NewBlockCipher};
 pub fn ige_encrypt(plaintext: &[u8], key: &[u8; 32], iv: &[u8; 32]) -> Vec<u8> {
     assert!(plaintext.len() % 16 == 0);
     let mut ciphertext = vec![0; plaintext.len()];
-    assert!(ciphertext.len() % 16 == 0);
```
Lonami, line 15: This was here because I thought it would've made the compiler happier (but maybe not).

Marekkon5: Yea, apparently doing own asserts makes the compiler not do them later on in the loops, but here it is redundant.

PR #86, opened 2021-08-25
Upload speed optimalizations

Marekkon5:
- SIMD in encryption (still not 100% like decryption, but faster than before)
- Directly generate uncompressed GZip for larger chunks
- Add connections parameter to upload_stream, instead of using global static

### DIFF
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -204,7 +204,7 @@ impl Client {
     /// // In-memory `Vec<u8>` buffers can be used as async streams
     /// let size = some_vec.len();
     /// let mut stream = std::io::Cursor::new(some_vec);
-    /// let uploaded_file = client.upload_stream(&mut stream, size, "sleep.jpg".to_string()).await?;
+    /// let uploaded_file = client.upload_stream(&mut stream, size, "sleep.jpg".to_string(), 4).await?;
     ///
     /// client.send_message(&chat, InputMessage::text("Zzz...").photo(uploaded_file)).await?;
     /// # Ok(())
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -217,7 +217,7 @@ impl Client {
         stream: &mut S,
         size: usize,
         name: String,
-        connections: usize
+        connections: usize,
     ) -> Result<Uploaded, io::Error> {
         let file_id = generate_random_id();
         let name = if name.is_empty() {
@@ -337,7 +337,8 @@ impl Client {
         // files, so it's fine to unwrap.
         let name = path.file_name().unwrap().to_string_lossy().to_string();
 
-        self.upload_stream(&mut file, size, name, WORKER_COUNT).await
+        self.upload_stream(&mut file, size, name, WORKER_COUNT)
+            .await
     }
 }
 
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -217,6 +217,7 @@ impl Client {
         stream: &mut S,
         size: usize,
         name: String,
+        connections: usize
     ) -> Result<Uploaded, io::Error> {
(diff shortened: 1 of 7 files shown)


Marekkon5:
Sorry for delays, did it the "proper way", set the default to None = always compress.

lib/grammers-client/src/client/files.rs
```diff
@@ -204,7 +204,7 @@ impl Client {
     /// // In-memory `Vec<u8>` buffers can be used as async streams
     /// let size = some_vec.len();
     /// let mut stream = std::io::Cursor::new(some_vec);
-    /// let uploaded_file = client.upload_stream(&mut stream, size, "sleep.jpg".to_string()).await?;
+    /// let uploaded_file = client.upload_stream(&mut stream, size, "sleep.jpg".to_string(), 4).await?;
```
Lonami, line 207: Eh we should figure out something nicer than this later on.

lib/grammers-crypto/src/lib.rs
```diff
@@ -115,10 +115,7 @@ fn do_encrypt_data_v2(plaintext: &[u8], auth_key: &AuthKey, random_padding: &[u8
         // "Note that MTProto 2.0 requires from 12 to 1024 bytes of padding"
         // "[...] the resulting message length be divisible by 16 bytes"
         let padding_len = determine_padding_v2_length(plaintext.len());
-        let mut buffer = Vec::with_capacity(plaintext.len() + padding_len);
-        buffer.extend(plaintext);
-        buffer.extend(&random_padding[..padding_len]);
-        buffer
+        vec![plaintext, &random_padding[..padding_len]].concat()
```
Lonami, line 118: This might be worth a comment, as I wouldn't expect this to have a large performance impact.

lib/grammers-mtproto/src/mtp/encrypted.rs
```diff
@@ -114,9 +116,15 @@ impl Builder {
         self
     }
 
-    /// Configures the compression threshold for outgoing messages.
-    pub fn compression_threshold(mut self, threshold: Option<usize>) -> Self {
-        self.compression_threshold = threshold;
+    /// Configures the min compression threshold for outgoing messages.
```
Lonami, line 119:
```suggestion
    /// Configures the lower bound of the compression threshold for outgoing messages.
```

Perhaps we should remove the `Option`? A value of `0` should do the same.

lib/grammers-mtproto/src/mtp/encrypted.rs
```diff
@@ -114,9 +116,15 @@ impl Builder {
...
+    pub fn min_compression_threshold(mut self, threshold: Option<usize>) -> Self {
+        self.min_compression_threshold = threshold;
+        self
+    }
+
+    /// Configures the max compression threshold for outgoing messages.
```
Lonami, line 125:
```suggestion
    /// Configures the upper bound of the compression threshold for outgoing messages.
    ///
    /// The message can be at most `threshold` bytes long for it to be compressed.
```

lib/grammers-mtproto/src/lib.rs
```diff
@@ -21,6 +21,7 @@ mod utils;
 
 /// The default compression threshold to be used.
 pub const DEFAULT_COMPRESSION_THRESHOLD: Option<usize> = Some(512);
+pub const DEFAULT_MAX_COMPRESSION_THRESHOLD: Option<usize> = None;
```
Lonami, line 24: Rename as `MIN` / `MAX`?

Lonami (changes requested):
> Sorry for delays, did it the "proper way", set the default to None = always compress.

I wanted to do it as well, but it looks like I've delayed even more.

Could you split the changes of compression and and encryption optimization into two though? Or rebase to have two proper commits so that I can merge the changes separatedly.

Lonami:
Gonna close this one since it's ancient. Can revisit in the future if there's interest.

PR #87, opened 2021-08-31
Encryption changes of PR #86

Marekkon5:
Just the encryption optimization part from the PR #86 .


lib/grammers-crypto/src/lib.rs
```diff
@@ -115,10 +115,8 @@ fn do_encrypt_data_v2(plaintext: &[u8], auth_key: &AuthKey, random_padding: &[u8
         // "Note that MTProto 2.0 requires from 12 to 1024 bytes of padding"
         // "[...] the resulting message length be divisible by 16 bytes"
         let padding_len = determine_padding_v2_length(plaintext.len());
-        let mut buffer = Vec::with_capacity(plaintext.len() + padding_len);
-        buffer.extend(plaintext);
-        buffer.extend(&random_padding[..padding_len]);
-        buffer
+        // `.concat()` is faster than .extend() according to flamegraph
```
Lonami, line 118:
```suggestion
        // `.concat()` is faster than `.extend()` according to flamegraph
```

PR #90, opened 2021-10-12
client: chats: mark builders as !Unpin

rupansh:
sry


Lonami:
What's the benefit to this?

PR #96, opened 2021-12-19
Added methods for joining and leaving a chat. Removed useless compile-time environment macros

BRA1L0R:
Join and leave methods:
Similar to the already implemented `kick_participant` which leaves a chat if you attempt to kick yourself, the now implemented `leave_chat` and `join_chat` follow the same structure. 

Remove of compile-time macro:
examples don't need compile-time environments to work. In fact, I think they have been misused in this case. I also fixed a clippy error in the process.

### DIFF
--- a/lib/grammers-client/examples/dialogs.rs
+++ b/lib/grammers-client/examples/dialogs.rs
@@ -14,7 +14,6 @@ use grammers_client::{Client, Config, SignInError};
 use grammers_session::Session;
 use log;
 use simple_logger::SimpleLogger;
-use std::env;
 use std::io::{self, BufRead as _, Write as _};
 use tokio::{runtime, task};
 
@@ -42,8 +41,8 @@ async fn async_main() -> Result<()> {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
 
     println!("Connecting to Telegram...");
     let mut client = Client::connect(Config {
--- a/lib/grammers-client/examples/echo.rs
+++ b/lib/grammers-client/examples/echo.rs
@@ -40,9 +40,9 @@ async fn async_main() -> Result {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
-    let token = env::args().skip(1).next().expect("token missing");
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
+    let token = env::args().nth(1).expect("token missing");
 
     println!("Connecting to Telegram...");
     let mut client = Client::connect(Config {
--- a/lib/grammers-client/examples/inline-pagination.rs
+++ b/lib/grammers-client/examples/inline-pagination.rs
@@ -110,9 +110,9 @@ async fn async_main() -> Result {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
-    let token = env::args().skip(1).next().expect("token missing");
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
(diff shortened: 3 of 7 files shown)


BRA1L0R:
fix: i removed the unused import that failed the build. my bad

lib/grammers-client/examples/dialogs.rs
```diff
@@ -42,8 +41,8 @@ async fn async_main() -> Result<()> {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
```
Lonami, line 45: `env!` is intentionally used because I think it's good to showcase that the **developer's** API ID and hash must be embedded in the binary (if you ship your application to a customer, they shouldn't have their own API ID and hash in the environment! Official applications also embed the value in the binary). So I would rather we removed this change.

lib/grammers-client/examples/echo.rs
```diff
@@ -40,9 +40,9 @@ async fn async_main() -> Result {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
-    let token = env::args().skip(1).next().expect("token missing");
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
```
Lonami, line 44: Same reasoning as above.

lib/grammers-client/examples/inline-pagination.rs
```diff
@@ -110,9 +110,9 @@ async fn async_main() -> Result {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
-    let token = env::args().skip(1).next().expect("token missing");
+    let api_id = std::env::var("TG_ID")?.parse()?;
+    let api_hash = std::env::var("TG_HASH")?;
```
Lonami, line 114: Same reasoning as above.

lib/grammers-client/src/client/chats.rs
```diff
@@ -406,6 +406,67 @@ impl Client {
...
+    }
+
+    /// Leaves a given chat
+    ///
+    /// Chat must be either a [`Channel`][chat] variant or a [`Group`][chat] variant.
+    ///
+    /// When used to join a [`User`][chat] variant an error is produced.
+    ///
+    /// Leaving a chat without being a participant will also result in an error.
+    ///
```
Lonami, line 445: This documentation should probably note that the chat is not deleted.

lib/grammers-client/src/types/chat/channel.rs
```diff
@@ -80,10 +80,17 @@ impl Channel {
     }
 
     /// Return the unique identifier for this channel.
+    #[inline]
```
Lonami, line 83: Not too happy about adding these changes in this PR.

lib/grammers-client/src/types/chat/channel.rs
```diff
@@ -80,10 +80,17 @@ impl Channel {
...
     pub fn id(&self) -> i32 {
         self.0.id
     }
 
+    /// Return access hash for this channel
+    #[inline]
+    pub fn access_hash(&self) -> Option<i64> {
```
Lonami, line 90: This probably should not be `pub`.

lib/grammers-client/src/types/chat/group.rs
```diff
@@ -67,16 +67,20 @@ impl Group {
         }
     }
 
+    /// Return the access hash for this group.
+    pub fn access_hash(&self) -> Option<i64> {
```
Lonami, line 71: Same as above (not `pub`).

lib/grammers-client/src/types/chat/mod.rs
```diff
@@ -79,6 +79,15 @@ impl Chat {
         }
     }
 
+    /// Return the access hash for this chat
+    pub fn access_hash(&self) -> Option<i64> {
```
Lonami, line 83: Same as above (not `pub`).

Lonami (changes requested):
Sorry that I took a long time to look into this. Unfortunately I would like to see some changes before merging.

Lonami:
Closing as this has been open for a long time with unresolved comments and no activity.

PR #97, opened 2021-12-23
Update to layer 135 & fixes

Manuel1510:
- Implement support for 64 bit chat/channel/user IDs and all layer 135 types.
- Fix a bug in grammers-tl-gen that would cause it to create an enum variant with an invalid identifier.
- Disable timestamp feature of simple_logger as it causes a panic on unix (due to time-rs not returning UTC offset, see https://github.com/time-rs/time/issues/380)


Lonami:
Thanks!

PR #98, opened 2021-12-28
add get_chat_members functions

tingfeng-key:


tingfeng-key:
I'm not familiar with this language. If there is a problem with the code, please give me guidance

lib/grammers-client/Cargo.toml
```diff
@@ -34,6 +34,7 @@ os_info = { version = "3.0.4", default_features = false }
 pin-project-lite = "0.2"
 pulldown-cmark = { version = "0.8.0", default-features = false, optional = true }
 tokio = { version = "1.5.0", features = ["sync", "fs", "macros", "time", "sync"] }
+anyhow = { version = "1.0.58" }
```
Lonami, line 37: So far the library has been doing a good job avoiding extra unnecessary dependencies. While this means a little more code to properly handle errors, I think it's worth it.

Lonami, line 37: So far the library has been doing a good job avoiding extra unnecessary dependencies. While this means a little more code to properly handle errors, I think it's worth it.

Cargo.toml
```diff
@@ -2,7 +2,6 @@
 members = [
     "bin/scrape-docs",
     "bin/tl-to-json",
-    "lib/grammers",
```
Lonami, line 5: Is this change needed for this PR?

lib/grammers-client/src/client/channel.rs
```diff
@@ -0,0 +1,75 @@
...
+    enums::{
+        channels::ChannelParticipants::{NotModified, Participants},
+        messages::ChatFull,
+        ChannelParticipantsFilter,
+        InputChannel::Channel,
+        User,
+    },
+    functions::channels::{GetFullChannel, GetParticipants},
+    types::InputChannel,
+};
```
Lonami, line 13: The library generally avoids this style of imports, so I think we should stick with flat `use`s for consistency.

lib/grammers/Cargo.toml
```diff
@@ -1,16 +0,0 @@
-[package]
```
Lonami, line 1: Why was the `grammers` module removed entirely?

lib/grammers-client/src/client/channel.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl Client {
+    const MAX_PARTICIPANT_LIMIT: usize = 200;
+
+    pub async fn get_full_channel(&mut self, channel_id: i64) -> Result<ChatFull> {
+        let input_channel = InputChannel {
+            channel_id,
+            access_hash: 0i64,
```
Lonami, line 21: This is unlikely to be reliable. The API generally expects you to know the access hash.

lib/grammers-client/src/client/channel.rs
```diff
@@ -0,0 +1,75 @@
...
+        User,
+    },
+    functions::channels::{GetFullChannel, GetParticipants},
+    types::InputChannel,
+};
+
+impl Client {
+    const MAX_PARTICIPANT_LIMIT: usize = 200;
+
+    pub async fn get_full_channel(&mut self, channel_id: i64) -> Result<ChatFull> {
```
Lonami, line 18: Instead of the `channel_id`, this should use the same `PackedChat` as input, like the other client methods do. See for example [`send_message`](https://github.com/Lonami/grammers/blob/3ed47232c01dc3c05426cc3c346e7deb1dadc002/lib/grammers-client/src/client/messages.rs#L381-L385) to know what the signature should look like.

lib/grammers-client/src/client/channel.rs
```diff
@@ -0,0 +1,75 @@
...
+            })
+            .await
+            .map_err(|e| anyhow!("get full channel error: {}", e.to_string()))?;
+
+        Ok(chat_full)
+    }
+
+    pub async fn get_chat_members(
+        &self,
+        channel_id: i64,
```
Lonami, line 35: Same rationale, please use `PackedChat`.

Lonami (changes requested):
Thanks for the PR. However, it needs quite a bit of work, style-wise and implementation-wise, there's several mismatches with my idea for grammers. Let me know if you need further clarification.

tingfeng-key:
Thank you very much for your advice. I will take the time to deal with these problems

PR #100, opened 2022-02-05
added offset for inline query

alenpaul2001:


Lonami:
The error is unrelated. Thanks!

PR #103, opened 2022-02-18
More friendly `Update` variants supported

George-Miao:
This is a WIP PR, I'll add more variants in future following this commit.


lib/grammers-client/src/types/update.rs
```diff
@@ -15,6 +15,8 @@ use std::sync::Arc;
 pub enum Update {
     /// Occurs whenever a new text message or a message with media is produced.
     NewMessage(Message),
+    /// Occurs when a message is updated.
+    MessageUpdated(Message),
```
Lonami, line 19: I think we should caml this `MessageEdited`, as per https://t.me/gramme_rs/11222.

George-Miao: Changed

George-Miao: > I think we should caml this `MessageEdited`, as per [t.me/gramme_rs/11222](https://t.me/gramme_rs/11222).

Changed

George-Miao:
Ready to be reviewed.

lib/grammers-client/src/types/update.rs
```diff
@@ -15,12 +15,19 @@ use std::sync::Arc;
...
+    /// Occurs when a message is deleted.
+    MessageDeleted(MessageDeletion),
     /// Occurs when Telegram calls back into your bot because an inline callback button was
     /// pressed.
     CallbackQuery(CallbackQuery),
     /// Occurs whenever you sign in as a bot and a user sends an inline query such as
     /// `@bot query`.
     InlineQuery(InlineQuery),
+    /// Raw events are not actual events.
+    /// Instead, they are the raw Update object that Telegram sends. You normally shouldn’t need these.
```
Lonami, line 29: Could we add a note like "the library can move events out of this variant into a new one across minor versions"? I don't mind the exact wording, but I want it to be clear that this variant pretty much means "WIP use this as a workaround".

lib/grammers-client/src/types/message_update.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Creates a new `MessageDeletion` from a vector of message IDs.
+    pub(crate) fn new(messages: Vec<i32>) -> Self {
+        Self {
+            channel_id: None,
+            messages,
+        }
+    }
+
+    /// Returns the channel ID if the message was deleted from a channel.
+    pub fn channel_id(&self) -> Option<i64> {
```
Lonami, line 29: I think this should do something similar to `Message::chat` (and maybe rename this to `chat` too).

https://github.com/Lonami/grammers/blob/master/lib/grammers-client/src/types/message.rs#L234-L240

I would like to hide away the raw API as much as possible where it makes sense. Users do not need to know only the channel ID is present (it's confusing, as some groups are channels too). Renaming and updating the documentation to say "this information is only known for megagroups and broadcast channels, and only the `id` is present" would be a better API IMO.

lib/grammers-client/src/types/update.rs
```diff
@@ -36,6 +38,9 @@ impl Update {
             tl::enums::Update::NewMessage(tl::types::UpdateNewMessage { message, .. }) => {
                 Message::new(client, message, chats).map(Self::NewMessage)
             }
+            tl::enums::Update::EditMessage(tl::types::UpdateEditMessage { message, .. }) => {
+                Message::new(client, message, chats).map(Self::NewMessage)
```
AndrielFR, line 44: you forgot to change from `Self::NewMessage` to `Self::MessageUpdated`

AndrielFR (changes requested):

lib/grammers-client/src/types/message_update.rs
George-Miao: Does this mean I should include a client inside the `MessageDeletion` or should that method accept a client as parameter? Since in my understanding this involves fetching data about the chat with the given `channel_id`?

PR #105, opened 2022-03-14
Implement `CallbackQuery::load_message`

AndrielFR:
So, you can use `CallbackQuery.load_message()`  to get the message from the pressed button.

### DIFF
--- a/lib/grammers-client/src/client/messages.rs
+++ b/lib/grammers-client/src/client/messages.rs
@@ -751,32 +751,6 @@ impl Client {
         Ok(message_ids.iter().map(|id| map.remove(id)).collect())
     }
 
-    /// Fetch a single message by ID.
-    ///
-    /// This is cheaper than [`Client::get_messages_by_id`].
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # async fn f(chat: grammers_client::types::Chat, mut client: grammers_client::Client) -> Result<(), Box<dyn std::error::Error>> {
-    /// let message_id = 123;
-    ///
-    /// let message = client.get_message_by_id(&chat, message_id).await?;
-    /// println!("{}", message.text());
-    /// # Ok(())
-    /// # }
-    /// ```
-    pub async fn get_message_by_id<C: Into<PackedChat>>(
-        &self,
-        chat: C,
-        message_id: i32,
-    ) -> Result<Option<Message>, InvocationError> {
-        let mut response = self.get_messages_by_id(chat, &[message_id]).await?;
-
-        Ok(Some(response.pop().unwrap().unwrap()))
-    }
-
-
     /// Get the latest pin from a chat.
     ///
     /// # Examples
--- a/lib/grammers-client/src/client/messages.rs
+++ b/lib/grammers-client/src/client/messages.rs
@@ -751,7 +751,9 @@ impl Client {
         Ok(message_ids.iter().map(|id| map.remove(id)).collect())
     }
 
-    /// Get a message using your ID.
+    /// Fetch a single message by ID.
+    ///
+    /// This is cheaper than [`Client::get_messages_by_id`].
     ///
     /// # Examples
     ///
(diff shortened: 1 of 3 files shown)


lib/grammers-client/src/client/messages.rs
```diff
@@ -751,6 +751,35 @@ impl Client {
         Ok(message_ids.iter().map(|id| map.remove(id)).collect())
     }
 
+    /// Get a message using your ID.
```
Lonami, line 754:
```suggestion
    /// Fetch a single message by ID.
```

If we could also say, "this is cheaper than `[get_messages]`" (linking to that function), I think that would be useful.

lib/grammers-client/src/client/messages.rs
```diff
@@ -751,6 +751,35 @@ impl Client {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # async fn f(chat: grammers_client::types::Chat, mut client: grammers_client::Client) -> Result<(), Box<dyn std::error::Error>> {
+    /// let message_id = 123;
+    ///
+    /// let message = client.get_message_by_id(&chat, message_id).await?;
```
Lonami, line 762: It's unfortunate that the API *technically* lets you fetch a message by ID without using the chat, but I guess it's fine to hide this from the user should that ever change.

lib/grammers-client/src/client/messages.rs
AndrielFR: oh, i think it would be good

Lonami:
I forgot we already had a version to get multiple messages by ID. Do you think adding a single-message variant is worthwhile? `.get(0)` on the other method would achieve this same result for a single ID.

Lonami:
> ok, i will use `.get(0)` then.

Sorry I didn't mean in this PR. I was asking if this PR is necessary given we already have a method which does something similar, and obtaining the behaviour from here is very easy with said existing method, making this addition more redundant.

AndrielFR:
> 

ok, do it your way, i just need `CallbackQuery::load_message` working.

lib/grammers-client/src/types/message.rs
```diff
@@ -476,10 +476,9 @@ impl Message {
         self.client
             .get_messages_by_id(&self.chat(), &[self.msg.id])
             .await?
-            .pop()
-            .unwrap()
+            .get(0)
```
Lonami, line 479: The error points this should've been kept unchanged.

lib/grammers-client/src/types/message.rs
```diff
@@ -476,10 +476,9 @@ impl Message {
...
             .unwrap();
-        todo!("actually mutate self after get_messages_by_id returns `Message`")
+        todo!("actually mutate self after get_message_by_id returns `Message`")
```
Lonami, line 481:
```suggestion
        todo!("actually mutate self after get_messages_by_id returns `Message`")
```

lib/grammers-client/src/types/callback_query.rs
```diff
@@ -77,7 +77,12 @@ impl CallbackQuery {
 
     /// Load the `Message` that contains the pressed inline button.
     pub async fn load_message(&self) -> Result<types::Message, InvocationError> {
-        todo!()
+        Ok(self
+            .client
+            .get_messages_by_id(self.chat(), &[self.query.msg_id])
+            .await?
+            .get(0)
```
Lonami, line 84: I think this will need to be a `pop` too if we want the function to return the owned type. When I mentioned `get(0)` I was thinking of user code. It was a brain fart.

lib/grammers-client/src/types/message.rs
AndrielFR: oh, ok

lib/grammers-client/src/types/callback_query.rs
AndrielFR: ok

lib/grammers-client/src/client/messages.rs
AndrielFR: I agree

PR #107, opened 2022-03-21
Add some usefull function for LoginToken

nrot:


nrot:
User login by code verification. I use finite-state machine for login procces. Becuase this is looped procces, i create login token var. But i can`t use without init. But i always know init state execute before use state. 
This is just example what i want to say:
```Rust
let mut state = State::GetToken;
let mut token = LoginToken::empty();
loop {
    match state{
        State::GetToken=>{
            token = client.request_login_code();
            state = State::CheckToken;
        },
        State::CheckToken=>{
            if let Ok(_) = client.sign_in(token){
                todo!();
            } else {
                token = LoginToken::empty();
                state = State::GetToken;
            }

        }
    }
}
```
_I use vscode formater and his has some different from cargo fmt_

Lonami:
I appreciate the pull request, but I do not think this addition would be a good idea. An empty token is not very useful, and Rust can represent empty values with the `Option` type. In my opinion, it's good design for a type to only represent valid values, and the empty case here is the odd one out. Your solution can use `Option` if it needs to (or better, have the token inside your `State`):

```rust
let mut state = State::GetToken;
let mut token = None; // 1
loop {
    match state{
        State::GetToken=>{
            token = Some(client.request_login_code()); // 2
            state = State::CheckToken;
        },
        State::CheckToken=>{
            if let Ok(_) = client.sign_in(token.unwrap()){ // 3
                todo!();
            } else {
                token = None; // 4
                state = State::GetToken;
            }

        }
    }
}
```

Feel free to join https://gramme_rs if you want to discuss other approaches.

nrot:
I tried to do it and it doesn't work.
```Rust 
Client::sign_in(&mut self, token: &LoginToken, code: &str)
``` 
This error from my code, but this still usefull for show problem
```
error[E0382]: use of moved value: `login_token`
   --> src\app.rs:205:48
    |
103 |         let mut login_token = None; //TIPS: This is just example
    |             --------------- move occurs because `login_token` has type `Option<LoginToken>`, which does not implement the `Copy` trait
...
149 |                             login_token = Some(s); //TIPS: Get LoginToken this
    |                             ----------- this reinitialization might get skipped
...
205 |                     match self.client.sign_in(&login_token.unwrap(), self.user_input_buf.as_str()).await {
    |                                                ^^^^^^^^^^^ -------- `login_token` moved due to this method call, in previous iteration of loop
    |
note: this function takes ownership of the receiver `self`, which moves `login_token`
   -->.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\option.rs:743:25
    |
743 |     pub const fn unwrap(self) -> T {
```
Reference can be moved by memory iter. Tried to copy unwrapped data to local variable. Just unwrap not help, because _`token` moved due to this method call, in previous iteration of loop_. I can`t add clone, because LoginToken not implement clone trait. 
```Rust
State::CheckToken=>{
    let lt = token.unwrap().clone();
    if let Ok(_) = client.sign_in(&lt){ // 3
        todo!();
    } else {
        token = None; // 4
        state = State::GetToken;
    }
}
```

Lonami:
`token.as_ref().unwrap()` should prevent the move.

PR #109, opened 2022-04-13
adds documentation for generated tl types

ehsan2003:
closes #108

PR #111, opened 2022-04-14
Fix examples env

### LINKED ISSUE #110: tests are failing due to examples env variable requirement
tests are failing due to missing environment variables in examples it might be better to have them as runtime dependencies instead of compile time. 

additionally having those critical information in a compiled code might be a security problem.

ehsan2003:
fixes #110


ehsan2003:
this closed due to my mistake of having unrelated commits ( related to #109 )

PR #112, opened 2022-04-21
Implement support to `MessageEntityMentionName`

AndrielFR:
With this, you can mention users either using html or markdown.

### DIFF
--- a/lib/grammers-client/src/parsers.rs
+++ b/lib/grammers-client/src/parsers.rs
@@ -11,6 +11,9 @@ use grammers_tl_types as tl;
 #[cfg(feature = "html")]
 const CODE_LANG_PREFIX: &str = "language-";
 
+#[cfg(any(feature = "markdown", feature = "html"))]
+const MENTION_URL_PREFIX: &str = "tg://user?id=";
+
 /// The length of a string, according to Telegram.
 ///
 /// Telegram considers the length of the string with surrogate pairs.
@@ -119,15 +122,15 @@ pub fn parse_markdown_message(message: &str) -> (String, Vec<tl::enums::MessageE
 
         // [text link](https://example.com) or [user mention](tg://user?id=12345678)
         Event::Start(Tag::Link(_kind, url, _title)) => {
-            if url.starts_with("tg://user?id") {
-                let user_id = url[13..].parse::<i64>().unwrap();
+            if url.starts_with(MENTION_URL_PREFIX) {
+                let user_id = url[MENTION_URL_PREFIX.len()..].parse::<i64>().unwrap();
                 push_entity!(MessageEntityMentionName(offset, user_id = user_id) => entities);
             } else {
                 push_entity!(MessageEntityTextUrl(offset, url = url.to_string()) => entities);
             }
         }
         Event::End(Tag::Link(_kindd, url, _title)) => {
-            if url.starts_with("tg://user?id") {
+            if url.starts_with(MENTION_URL_PREFIX) {
                 update_entity_len!(MentionName(offset) => entities);
             } else {
                 update_entity_len!(TextUrl(offset) => entities);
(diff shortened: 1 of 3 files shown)


lib/grammers-client/src/parsers.rs
```diff
@@ -117,12 +117,21 @@ pub fn parse_markdown_message(message: &str) -> (String, Vec<tl::enums::MessageE
             update_entity_len!(Italic(offset) => entities);
         }
 
-        // [text link](https://example.com)
+        // [text link](https://example.com) or [user mention](tg://user?id=12345678)
         Event::Start(Tag::Link(_kind, url, _title)) => {
-            push_entity!(MessageEntityTextUrl(offset, url = url.to_string()) => entities);
+            if url.starts_with("tg://user?id") {
```
Lonami, line 122: Please refactor this into a `const MENTION_URL_PREFIX` (name can be better).

lib/grammers-client/src/parsers.rs
```diff
@@ -117,12 +117,21 @@ pub fn parse_markdown_message(message: &str) -> (String, Vec<tl::enums::MessageE
...
+                let user_id = url[13..].parse::<i64>().unwrap();
```
Lonami, line 123: Once we have the const, use its `.len()`.

lib/grammers-client/src/parsers.rs
```diff
@@ -117,12 +117,21 @@ pub fn parse_markdown_message(message: &str) -> (String, Vec<tl::enums::MessageE
...
+                push_entity!(MessageEntityMentionName(offset, user_id = user_id) => entities);
+            } else {
+                push_entity!(MessageEntityTextUrl(offset, url = url.to_string()) => entities);
+            }
         }
-        Event::End(Tag::Link(_kindd, _url, _title)) => {
-            update_entity_len!(TextUrl(offset) => entities);
+        Event::End(Tag::Link(_kindd, url, _title)) => {
+            if url.starts_with("tg://user?id") {
```
Lonami, line 130: Same here.

lib/grammers-client/src/parsers.rs
```diff
@@ -237,8 +246,14 @@ pub fn parse_html_message(message: &str) -> (String, Vec<tl::enums::MessageEntit
                             .map(|a| a.value.to_string())
                             .unwrap_or_else(|| "".to_string());
 
-                        push_entity!(MessageEntityTextUrl(self.offset, url = url)
-                            => self.entities);
+                        if url.starts_with("tg://user?id") {
```
Lonami, line 249: Same `const`.

lib/grammers-client/src/parsers.rs
```diff
@@ -237,8 +246,14 @@ pub fn parse_html_message(message: &str) -> (String, Vec<tl::enums::MessageEntit
...
+                            let user_id = url[13..].parse::<i64>().unwrap();
```
Lonami, line 250: `const`'s `.len()`.

lib/grammers-client/src/parsers.rs
AndrielFR: ok.

Lonami (commented):
The current CI error is unrelated.

The next step is to change `Client::send_message` to look for the message entities, replacing all these `MessageEntityMentionName` with [`InputMessageEntityMentionName`](https://docs.rs/grammers-tl-types/latest/grammers_tl_types/enums/enum.MessageEntity.html#variant.InputMessageEntityMentionName).

Lonami:
Just to confirm, if you use this code to send a message with a mention, the mention is removed by the server, right? I don't think things have changed since I tried, but maybe.

AndrielFR:
> Just to confirm, if you use this code to send a message with a mention, the mention is removed by the server, right? I don't think things have changed since I tried, but maybe.

yes, i can change all to `InputMessageEntityMention` to avoid it?

Lonami:
Yes that's the next step (I detailed it some more in the Telegram group).

AndrielFR:
> Yes that's the next step (I detailed it some more in the Telegram group).

i'm doing the function to parse it.

Lonami (changes requested):
We can merge this once we document this special link in https://github.com/Lonami/grammers/blob/4e95ca5e37fc6019b06b73b9d2dade9b8561e9cd/lib/grammers-client/src/types/input_message.rs#L313 (and the `html`). Something like "inline links with this format (…) will be replaced with inline mentions when possible".

AndrielFR:
> We can merge this once we document this special link in
> 
> https://github.com/Lonami/grammers/blob/4e95ca5e37fc6019b06b73b9d2dade9b8561e9cd/lib/grammers-client/src/types/input_message.rs#L313
> (and the `html`). Something like "inline links with this format (…) will be replaced with inline mentions when possible".

the way i did is ok?

Lonami:
Yes, docs are fine, but `rustfmt` is complaining.

AndrielFR:
> Yes, docs are fine, but `rustfmt` is complaining.

done.

Lonami:
Thank you!

PR #118, opened 2022-07-11
Port messagebox from Telethon

mkpankov:
Each commit that is a direct copy contains hash of commit in Telethon.

Those not having it are fixups that don't directly correspond to any Telethon commits.

### DIFF
--- a/lib/grammers-session/src/message_box/mod.rs
+++ b/lib/grammers-session/src/message_box/mod.rs
@@ -421,6 +421,11 @@ impl MessageBox {
         update: tl::enums::Update,
         reset_deadline: ResetDeadline,
     ) -> Option<tl::enums::Update> {
+        if let tl::enums::Update::ChannelTooLong(u) = update {
+            self.begin_get_diff(Entry::Channel(u.channel_id));
+            return None;
+        }
+
         let pts = match PtsInfo::from_update(&update) {
             Some(pts) => pts,
             // No pts means that the update can be applied in any order.
--- a/lib/grammers-session/src/message_box/mod.rs
+++ b/lib/grammers-session/src/message_box/mod.rs
@@ -536,7 +536,10 @@ impl MessageBox {
         for entry in [Entry::AccountWide, Entry::SecretChats] {
             if self.getting_diff_for.contains(&entry) {
                 if self.map.get(&entry).is_none() {
-                    panic!("Should not try to get difference for an entry {:?} without known state", entry);
+                    panic!(
+                        "Should not try to get difference for an entry {:?} without known state",
+                        entry
+                    );
                 }
 
                 return Some(tl::functions::updates::GetDifference {
@@ -742,13 +745,10 @@ impl MessageBox {
                     },
                 })
             } else {
-                // TODO investigate when/why/if this can happen
-                warn!(
-                    "cannot getChannelDifference for {} as we're missing its pts",
-                    id
+                panic!(
+                    "Should not try to get difference for an entry {:?} without known state",
+                    entry
                 );
-                self.end_get_diff(entry);
-                None
             }
         } else {
             warn!(
--- a/lib/grammers-session/src/message_box/mod.rs
+++ b/lib/grammers-session/src/message_box/mod.rs
@@ -808,7 +808,7 @@ impl MessageBox {
 
                 self.map.get_mut(&entry).unwrap().pts = pts;
                 updates.extend(new_messages.into_iter().map(|message| {
-                    tl::types::UpdateNewMessage {
+                    tl::types::UpdateNewChannelMessage {
                         message,
                         pts: NO_SEQ,
                         pts_count: NO_SEQ,
--- a/lib/grammers-session/src/message_box/mod.rs
+++ b/lib/grammers-session/src/message_box/mod.rs
@@ -488,7 +488,7 @@ impl MessageBox {
         } else {
             // No previous `pts` known, and because this update has to be "right" (it's the first one) our
             // `local_pts` must be one less.
-            pts.pts - 1
+            pts.pts - pts.pts_count
         };
 
         // For example, when we're in a channel, we immediately receive:
--- a/lib/grammers-session/src/message_box/mod.rs
+++ b/lib/grammers-session/src/message_box/mod.rs
@@ -197,8 +197,7 @@ impl MessageBox {
             state.deadline = deadline;
             debug!("reset deadline {:?} for {:?}", deadline, entry);
         } else {
-            // TODO figure out why this happens
-            info!("did not reset deadline for {:?} as it had no entry", entry);
+            panic!("did not reset deadline for {:?} as it had no entry", entry);
         }
 
         if self.next_deadline == Some(entry) {
@@ -448,6 +447,8 @@ impl MessageBox {
         // the "no updates" period for that entry is reset.
         //
         // Build the `HashSet` to avoid calling `reset_deadline` more than once for the same entry.
+        //
+        // By the time this method returns, self.map will have an entry for which we can reset its deadline.
         if reset_deadline == ResetDeadline::Yes {
             self.reset_deadlines_for.insert(pts.entry);
         }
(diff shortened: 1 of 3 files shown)


lib/grammers-session/src/message_box/mod.rs
```diff
@@ -278,10 +277,22 @@ impl MessageBox {
...
     }
 
-    /// Begin getting difference for the given entry.
+    /// Try to begin getting difference for the given entry.
+    /// Fails if the entry does not have a previously-known state that can be used to get its difference.
     ///
     /// Clears any previous gaps.
-    fn begin_get_diff(&mut self, entry: Entry) {
+    fn try_begin_get_diff(&mut self, entry: Entry) {
+        if self.map.get(&entry).is_none() {
```
Lonami, line 285: I'm guessing this and the check below can both use [`contains_key`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html) (but it's fine to leave it as-is; both should work the same).

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -511,19 +532,25 @@ impl MessageBox {
 impl MessageBox {
     /// Return the request that needs to be made to get the difference, if any.
     pub fn get_difference(&mut self) -> Option<tl::functions::updates::GetDifference> {
-        let entry = Entry::AccountWide;
-        if self.getting_diff_for.contains(&entry) {
-            if let Some(state) = self.map.get(&entry) {
+        for entry in [Entry::AccountWide, Entry::SecretChats] {
+            if self.getting_diff_for.contains(&entry) {
+                if self.map.get(&entry).is_none() {
```
Lonami, line 537: Same here for the `contains_key`.

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -568,59 +601,88 @@ impl MessageBox {
...
+        // We need to process those as if they were socket updates to discard any we have already handled.
+        let mut result_updates = vec![];
+        let us = tl::enums::Updates::Updates(tl::types::Updates {
+            updates,
+            users: users.clone(),
+            chats: chats.clone(),
+            date: 1,
+            seq: NO_SEQ,
         });
+        self.process_updates(us, chat_hashes, &mut result_updates)?;
```
Lonami, line 683: In theory, applying the updates from `getDifference` should **not** trigger another gap (that's the very thing `getDifference` is supposed to be used for!).

I see that the client was updated to log an error if for some reason this does happen. Personally I'd be more inclined to panic here (the server must be misbehaving if it's returning gaps from `getDifference`), but I guess we can leave it to log an error for now and we'll see if it ever happens.

I'm also not sure how the client should behave if this does happen. I'm inclined to panic because we're probably not capable of handling `getDifference` returning gaps.

lib/grammers-session/src/message_box/mod.rs
mkpankov: Done

lib/grammers-session/src/message_box/mod.rs
mkpankov: Done

lib/grammers-session/src/message_box/mod.rs
mkpankov: Done

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -282,7 +282,7 @@ impl MessageBox {
     ///
     /// Clears any previous gaps.
     fn try_begin_get_diff(&mut self, entry: Entry) {
-        if self.map.get(&entry).is_none() {
+        if self.map.contains_key(&entry) {
```
Lonami, line 285: This would have to be:
```rust
         if !self.map.contains_key(&entry) {
         // ^
```

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -282,7 +282,7 @@ impl MessageBox {
...
             // Won't actually be able to get difference for this entry if we don't have a pts to start off from.
             if self.possible_gaps.get(&entry).is_some() {
```
Lonami, line 287: This wasn't changed to `contains_key`.

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -344,8 +344,7 @@ impl MessageBox {
         } = match adaptor::adapt(updates, chat_hashes) {
             Ok(combined) => combined,
             Err(Gap) => {
-                self.try_begin_get_diff(Entry::AccountWide);
-                return Err(Gap);
+                panic!("adaptor::adapt should not trigger another gap");
```
Lonami, line 347: It's okay to trigger a gap when we process socket updates. What I meant by my previous comment was that we should panic if the gap is detected *while applying difference* (changing the propagation of the gap in the method that applies difference to an `expect`). But we should not panic here at the root of the error (because this code is also used to process socket updates which are expected to trigger gaps).

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -680,7 +677,7 @@ impl MessageBox {
             date: 1,
             seq: NO_SEQ,
         });
-        self.process_updates(us, chat_hashes, &mut result_updates)?;
+        self.process_updates(us, chat_hashes, &mut result_updates);
```
Lonami, line 680: I meant we should `.expect` here.

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -847,7 +844,7 @@ impl MessageBox {
                     date: 1,
                     seq: NO_SEQ,
                 });
-                self.process_updates(us, chat_hashes, &mut result_updates)?;
+                self.process_updates(us, chat_hashes, &mut result_updates);
```
Lonami, line 847: ...and `.expect` here.

mkpankov:
Updated

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -278,10 +277,22 @@ impl MessageBox {
...
     /// Clears any previous gaps.
-    fn begin_get_diff(&mut self, entry: Entry) {
+    fn try_begin_get_diff(&mut self, entry: Entry) {
+        if !self.map.contains_key(&entry) {
+            // Won't actually be able to get difference for this entry if we don't have a pts to start off from.
+            if self.possible_gaps.contains_key(&entry) {
+                log::error!(
+                    "Should not have a possible_gap for an entry {:?} not in the state map",
+                    entry
+                );
```
Lonami, line 291: I think this was intended to be a `panic!()`.

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -534,30 +561,36 @@ impl MessageBox {
         &mut self,
         difference: tl::enums::updates::Difference,
         chat_hashes: &mut ChatHashCache,
-    ) -> (
-        Vec<tl::enums::Update>,
-        Vec<tl::enums::User>,
-        Vec<tl::enums::Chat>,
-    ) {
+    ) -> Result<
```
Lonami, line 564: It seems we always return `Ok`, so this `Result<>`  is redundant (unless I missed a `?`).

lib/grammers-session/src/message_box/mod.rs
```diff
@@ -707,11 +764,14 @@ impl MessageBox {
         request: tl::functions::updates::GetChannelDifference,
         difference: tl::enums::updates::ChannelDifference,
         chat_hashes: &mut ChatHashCache,
-    ) -> (
-        Vec<tl::enums::Update>,
-        Vec<tl::enums::User>,
-        Vec<tl::enums::Chat>,
-    ) {
+    ) -> Result<
```
Lonami, line 767: Same here, we always return `Ok` now, so the change to return `Result` is not needed.

Lonami:
Please run `cargo fmt`.

mkpankov:
Done

PR #123, opened 2022-07-21
Cargo clippy fixes, and edition upgrade

lorislibralato:
I followed `cargo clippy` suggestions and modified the code manually.
This is my first open source PR, if anything is wrong I'm happy to fix.

### DIFF
--- a/lib/grammers-tl-gen/src/enums.rs
+++ b/lib/grammers-tl-gen/src/enums.rs
@@ -17,7 +17,7 @@ use std::collections::HashSet;
 use std::io::{self, Write};
 
 /// Types that implement Copy from builtin_type
-const COPY_TYPES: [&'static str; 5] = ["bool", "f64", "i32", "i64", "u32"];
+const COPY_TYPES: [&str; 7] = ["bool", "f64", "i32", "i64", "u32", "[u8; 16]", "[u8; 32]"];
 
 /// Writes an enumeration listing all types such as the following rust code:
 ///
@@ -63,7 +63,7 @@ fn write_enum<W: Write>(
         if metadata.is_recursive_def(d) {
             write!(file, "Box<")?;
         }
-        write!(file, "{}", rustifier::definitions::qual_name(&d))?;
+        write!(file, "{}", rustifier::definitions::qual_name(d))?;
         if metadata.is_recursive_def(d) {
             write!(file, ">")?;
         }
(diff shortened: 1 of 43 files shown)


alenpaul2001:
changing the edition necessary ?

lib/grammers-tl-gen/src/enums.rs
```diff
@@ -17,7 +17,7 @@ use std::collections::HashSet;
 use std::io::{self, Write};
 
 /// Types that implement Copy from builtin_type
-const COPY_TYPES: [&'static str; 5] = ["bool", "f64", "i32", "i64", "u32"];
+const COPY_TYPES: [&str; 7] = ["bool", "f64", "i32", "i64", "u32", "[u8; 16]", "[u8; 32]"];
```
Lonami, line 20: This is a behavior change. Do we need to do this in this PR?

lib/grammers-tl-gen/src/rustifier.rs
```diff
@@ -98,20 +98,20 @@ pub mod definitions {
...
                 &name[name
                     .as_bytes()
                     .iter()
+                    .take(name.len() - variant.len())
                     .rposition(|c| c.is_ascii_uppercase())
                     .unwrap_or(0)..]
             }
-            "Self" => {
-                // Use the name from the second-to-last uppercase letter
+            _ if variant.chars().all(char::is_numeric) => {
```
Lonami, line 110: Curious to know why this got re-ordered.

Lonami:
> changing the edition necessary ?

I'm personally okay with it. I don't think there's a MSRV we should worry about. Not that we were enforcing one before anyway.

lib/grammers-tl-gen/src/enums.rs
lorislibralato: In generated.rs int128 and int256 used .clone() and this Is unecessary because they implement Copy

lib/grammers-tl-gen/src/rustifier.rs
lorislibralato: Clippy warn the case `"" | _` so I changed the order of match the cases and used only `_`

lib/grammers-tl-gen/src/enums.rs
Lonami: Ah, so clippy was complaining about the generated code. That's fair.

lib/grammers-tl-gen/src/rustifier.rs
Lonami: I remember now, I left the explicit `""` arm because it's not obvious that we also want to match empty strings with `all`. Can we add a comment?

```rust
// Note: this intentionally handles the empty string case too.
```

lorislibralato: Done!

Lonami:
Thanks.

Issue #126, opened 2022-08-02
Grammers-client socks5 proxy feature request


niuhuan:
My country does not allow me to directly access telegram server.

Even though it can be deployed to servers in other countries, there are still problems in the development process

I want to use Socks5 or HTTP proxy to access the server

Lonami:
I am personally not interested in dealing with proxies. It might be easier to set a system-wide proxy. If however you are willing to contribute support for them (and the code is reasonably simple to maintain), I will happily merge such a feature.

niuhuan:
I'm so sorry because I'm a novice in rust. 

I tried proxychans, set environments, set `tl::functions::InitConnection` proxy field, but cannot use proxy both.

Can you outline the way to implement the proxy?

And thanks.

Lonami:
To be fair, I have no idea what the proxy field of `InitConnection` does. This was a latter addition from Telegram. From what I remember when I merged proxy support in Telethon, all that was done was add a dependency on a library capable of connecting through SOCKS5 proxies. I'm assuming we would need to find something similar in crates.io and use that in the grammers-mtsender lib.

PR #127, opened 2022-08-05
connect_with_proxy

niuhuan:
#126  

Add proxy field to InitParms. Connect server with proxy if set.

### DIFF
--- a/lib/grammers-client/DEPS.md
+++ b/lib/grammers-client/DEPS.md
@@ -79,3 +79,19 @@ without having to use `Box`.
 ## futures-util
 
 Provides useful functions for working with futures/tasks.
+
+## url
+
+Parse proxy url
+
+## trust-dns-resolver
+
+Lookup ip address if proxy host is domain
+
+## tokio-socks
+
+Socks5 proxy supported
+
+## rand
+
+Choose ip if proxy domain has any address
--- a/lib/grammers-client/DEPS.md
+++ b/lib/grammers-client/DEPS.md
@@ -79,19 +79,3 @@ without having to use `Box`.
 ## futures-util
 
 Provides useful functions for working with futures/tasks.
-
-## url
-
-Parse proxy url
-
-## trust-dns-resolver
-
-Lookup ip address if proxy host is domain
-
-## tokio-socks
-
-Socks5 proxy supported
-
-## rand
-
-Choose ip if proxy domain has any address
--- a/lib/grammers-client/src/client/client.rs
+++ b/lib/grammers-client/src/client/client.rs
@@ -92,6 +92,8 @@ pub struct InitParams {
     ///
     /// When the limit is `Some`, a buffer to hold that many updates will be pre-allocated.
     pub update_queue_limit: Option<usize>,
+    /// Proxy url of TcpStream
+    pub proxy: Option<String>,
 }
 
 pub(crate) struct ClientInner {
@@ -154,6 +156,7 @@ impl Default for InitParams {
             server_addr: None,
             flood_sleep_threshold: Some(60),
             update_queue_limit: Some(100),
+            proxy: None,
         }
     }
 }
--- a/lib/grammers-client/src/client/client.rs
+++ b/lib/grammers-client/src/client/client.rs
@@ -92,7 +92,11 @@ pub struct InitParams {
     ///
     /// When the limit is `Some`, a buffer to hold that many updates will be pre-allocated.
     pub update_queue_limit: Option<usize>,
-    /// Proxy url of TcpStream
+    /// URL of the proxy to use.
+    ///
+    /// The scheme must be `socks5`. Username and password are optional.
+    ///
+    /// Both a host and port must be provided. If a domain is used for the host, domain, its address will be looked up.
     pub proxy: Option<String>,
 }
 
--- a/lib/grammers-mtsender/DEPS.md
+++ b/lib/grammers-mtsender/DEPS.md
@@ -30,18 +30,18 @@ Used in the tests in order to debug with more information when things go wrong.
 
 Used to test that this file lists all dependencies from `Cargo.toml`.
 
-## tokio-socks
-
-Socks5 proxy support.
-
 ## url
 
-Parse proxy url.
+Used to parse the optional proxy URL.
 
 ## trust-dns-resolver
 
-Lookup ip if proxy host is a domain.
+Used to look up the IP address of the proxy host if a domain is provided.
+
+## tokio-socks
+
+SOCKS5 proxy support.
 
 ## rand
 
-Random choose ip from domain.
+Used to choose one of the many IP addresses of the proxy's domain.
--- a/lib/grammers-mtsender/DEPS.md
+++ b/lib/grammers-mtsender/DEPS.md
@@ -29,3 +29,19 @@ Used in the tests in order to debug with more information when things go wrong.
 ## toml
 
 Used to test that this file lists all dependencies from `Cargo.toml`.
+
+## tokio-socks
+
+Socks5 proxy support.
+
+## url
+
+Parse proxy url.
+
+## trust-dns-resolver
+
+Lookup ip if proxy host is a domain.
+
+## rand
+
+Random choose ip from domain.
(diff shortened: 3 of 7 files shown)


lib/grammers-mtsender/src/lib.rs
```diff
@@ -1,29 +1,37 @@
-// Copyright 2020 - developers of the `grammers` project.
-//
-// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
-// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
-// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
-// option. This file may not be copied, modified, or distributed
-// except according to those terms.
-mod errors;
-
```
Lonami, line 9: This should not be removed.

lib/grammers-mtsender/src/lib.rs
```diff
@@ -68,7 +76,7 @@ pub(crate) fn generate_random_id() -> i64 {
 
 // Manages enqueuing requests, matching them to their response, and IO.
 pub struct Sender<T: Transport, M: Mtp> {
-    stream: TcpStream,
+    stream: Box<dyn Stream>,
```
Lonami, line 79: This is an unfortunate (tiny) performance hit for the use-case where no proxies are used, but that can be solved with a new generic parameter.

lib/grammers-mtsender/src/lib.rs
```diff
@@ -128,13 +136,74 @@ impl Enqueuer {
 }
 
 impl<T: Transport, M: Mtp> Sender<T, M> {
-    async fn connect<A: ToSocketAddrs>(
+    async fn connect(
         transport: T,
         mtp: M,
-        addr: A,
+        addr: SocketAddr,
```
Lonami, line 142: Why was this changed from the generic parameter?

lib/grammers-mtsender/src/lib.rs
```diff
@@ -128,13 +136,74 @@ impl Enqueuer {
...
     ) -> Result<(Self, Enqueuer), io::Error> {
         info!("connecting...");
-        let stream = TcpStream::connect(addr).await?;
+
+        let stream: Box<dyn Stream> = match std::env::var("GRAMMERS_CLIENT_PROXY") {
```
Lonami, line 146: This will need to change before I can merge it. I propose we have a separate method, called something like `connect_with_proxy`, which is just like `connect`, but also accepts the proxy URL.

lib/grammers-mtsender/src/lib.rs
```diff
@@ -493,9 +562,9 @@ impl<T: Transport> Sender<T, mtp::Encrypted> {
     }
 }
 
-pub async fn connect<T: Transport, A: ToSocketAddrs>(
+pub async fn connect<T: Transport>(
     transport: T,
-    addr: A,
+    addr: SocketAddr,
```
Lonami, line 567: Why was the generic removed? (Same as before.)

lib/grammers-mtsender/src/lib.rs
```diff
@@ -540,10 +609,26 @@ pub async fn connect<T: Transport, A: ToSocketAddrs>(
     ))
 }
 
-pub async fn connect_with_auth<T: Transport, A: ToSocketAddrs>(
+pub async fn connect_with_auth<T: Transport>(
     transport: T,
-    addr: A,
+    addr: SocketAddr,
```
Lonami, line 614: Why was the generic removed? (Same as before.)

Lonami (changes requested):
Thanks for the PR. If this solves your issues, I'm happy to merge, once we resolve the problem around using the environment to decide whether or not to use a proxy, and the other small nits.

Lonami:
Note that while I'm OK with this code being in the main repository I still have no plans to directly support it (that will probably still be on the community). I just have no personal interest in dealing with proxies.

niuhuan:
Thanks code review, this taught me a lot of knowledge , I will try to modify it according to your ideas.

lib/grammers-mtsender/src/lib.rs
niuhuan: Sorry, I'not found way convert ToSocketAddress to ToTargetAddress, I want it to run first

niuhuan:
Had modify to `bot_sign_in_with_proxy`,  `request_login_code_with_proxy`, `connect_with_proxy`

lib/grammers-mtsender/src/lib.rs
niuhuan: Do you have any suggestions here?

niuhuan:
I think must be known type befor use generic .

And this really bothered me. 

捕获

niuhuan:
If anyone has spare time, I hope to solve this problem ( Make the code look better ). Proxy is too important for Chinese Mainland.

lib/grammers-mtsender/src/lib.rs
Lonami: No, it's fine to leave this as-is for now.

Lonami (changes requested):
I hadn't thought about this before, but after seeing the changes to add `bot_sign_in_with_proxy` and `request_login_code_with_proxy`, I am confident that the right solution would be to add:

```rust
proxy_url: Option<String>,
```

to [`InitParams`](https://github.com/Lonami/grammers/blob/ad3fd4b7facafccc1eeacbf1df11a67b58d52e60/lib/grammers-client/src/client/client.rs#L46). This way, we don't need to duplicate every login and connect method, and it remains optional.

So my idea is:

* Revert the addition of `bot_sign_in_with_proxy`, `request_login_code_with_proxy`, and `connect_with_proxy` in the `client`.
* Remove `proxy: Option<String>` from `connect_sender` in the `client`.

I'm okay with keeping the `proxy` parameter in the `mtsender` crate for the time being (that's not really public API I care much about for the time being).

We can also merge the removal of the generic `A: ToSocketAddrs` for the time being, but I'd like to open an issue after we merge so we don't forget to fix that. This does not need to be fixed in this PR.

niuhuan:
I' had modify `SocksAddr` to `ToSocksAddr + IntoTargetAddr`, It looks good now.

niuhuan:
DEPS.md has been update

lib/grammers-client/src/client/client.rs
```diff
@@ -92,6 +92,8 @@ pub struct InitParams {
     ///
     /// When the limit is `Some`, a buffer to hold that many updates will be pre-allocated.
     pub update_queue_limit: Option<usize>,
+    /// Proxy url of TcpStream
```
Lonami, line 95:
```suggestion
    /// URL of the proxy to use.
    ///
    /// The scheme must be `socks5`. Username and password are optional.
    ///
    /// Both a host and port must be provided. If a domain is used for the host, domain, its address will be looked up.
```

lib/grammers-mtsender/DEPS.md
```diff
@@ -29,3 +29,19 @@ Used in the tests in order to debug with more information when things go wrong.
...
+
+Parse proxy url.
+
+## trust-dns-resolver
+
+Lookup ip if proxy host is a domain.
+
+## rand
+
+Random choose ip from domain.
```
Lonami, line 47:
```suggestion
## url

Used to parse the optional proxy URL.

## trust-dns-resolver

Used to look up the IP address of the proxy host if a domain is provided.

## tokio-socks

SOCKS5 proxy support.

## rand

Used to choose one of the many IP addresses of the proxy's domain.
```

lib/grammers-client/DEPS.md
```diff
@@ -79,3 +79,19 @@ without having to use `Box`.
...
+
+Lookup ip address if proxy host is domain
+
+## tokio-socks
+
+Socks5 proxy supported
+
+## rand
+
+Choose ip if proxy domain has any address
```
Lonami, line 97: `grammers-client` does not need to be updated because its `Cargo.toml` does not depend on these so we don't need to update this file.

```suggestion
```

Lonami (changes requested):

Lonami (approved):

Lonami:
Thanks. I still plan to make further changes but we can merge it for now.

Lonami:
Hi @niuhuan, would you mind testing the above commit? Notably:

* I have made the proxy feature optional. You must enable it in order to specify a proxy.
* I have removed the dependency on `rand`. I don't think randomness is a good fit for this task.
* I have renamed the field to `proxy_url` because I think that's clearer for the end user.

I've also changed a bit how the code is structured. It should still work fine, but I would like it if you could confirm if these changes work for you.

niuhuan:
That's good, I had test on my windows desktop, It's running. Thanks

PR #139, opened 2022-09-26
Implement `Channel::username()`

AndrielFR:
With this, you can use `channel.username()` to get a megagroup/channel username.


Lonami:
The current CI error is unrelated to this change. However I'd like to see this method added for `Chat` and `Group` as well before merging (`User` already has it). Would you be willing to add these?

AndrielFR:
> The current CI error is unrelated to this change. However I'd like to see this method added for `Chat` and `Group` as well before merging (`User` already has it). Would you be willing to add these?

Done. (?)

Lonami:
Excellent, thanks!

AndrielFR:
> Excellent, thanks!

🤝

Issue #135, opened 2022-08-26
Packed is invalid after restart


tingfeng-key:
The code for getting packed is as follows:
```rust
while let Some(update) = self.client.next_update().await? {
      if let Update::NewMessage(message) = update {
          if let Some(Chat::User(types::User(user))) = message.sender() {
              warn!(
                  "{}",
                  serde_json::to_string(
                      &Chat::from_user(user.into()).pack().to_bytes()
                  )
                  .unwrap()
              );
          }
    }
```
After printing the packed string, stop the program and execute the following code:
```rust
let encode: Vec<u8> = vec![];//Real packed.to_ Bytes data

let decode = PackedChat::from_bytes(&encode).expect("error");

let input_user = decode.try_to_input_user().expect("input error");

//get_full_user is the method that implements tl::functions::users::GetFullUser
let full_user = self.client.clone().get_full_user(input_user.clone()).await;

warn!("{:#?}", full_user);

//get_users is the method that implements tl::functions::users::GetUsers
let users = self.client.clone().get_users(vec![input_user]).await;
warn!("{:#?}", users);

let msg = InputMessage::text("hi");
let m = self.client.send_message(decode, msg).await;
warn!("{:#?}", m);
```
After running the program, it prints as follows：
```rust
Err(
    Other(
        Rpc(
            RpcError {
                code: 400,
                name: "USER_ID_INVALID",
                value: None,
                caused_by: Some(
                    3054459160,
                ),
            },
        ),
    ),
)
Ok(
    [],
)
Err(
    Rpc(
        RpcError {
            code: 400,
            name: "PEER_ID_INVALID",
            value: None,
            caused_by: Some(
                228423076,
            ),
        },
    ),
)
```
The implementation of get_users and get_full_user is as follows：
```rust
    pub async fn get_full_user(
        &mut self,
        id: tl::enums::InputUser,
    ) -> Result<(tl::types::UserFull, tl::types::User), UserError> {
        match self.invoke(&tl::functions::users::GetFullUser { id }).await {
            Ok(tl::enums::users::UserFull::Full(user_full)) => {
                let full_user = match user_full.full_user {
                    tl::enums::UserFull::Full(user) => user,
                };

                let user_base = match user_full.users.first() {
                    Some(tl::enums::User::Empty(_)) => Err(UserError::EmptyUser),
                    Some(tl::enums::User::User(user)) => Ok(user.clone()),
                    None => Err(UserError::EmptyUser),
                }?;
[...]

PR #144, opened 2022-09-27
add username/access_hash/is_min method

tingfeng-key:


Lonami (commented):
Thanks for the PR. However I would prefer to not expose things like `access_hash` or `min` directly to the users. These are mostly "implementation details" when using Telegram's API. (I'm okay with adding `username` but #139 already does that).

What need do you have for exposing `min` and `access_hash`? We can discuss this and maybe come to a middle ground where we're both happy.

tingfeng-key:
OK
`min` is due to this problem(#135 ). I need to filter out `access_hash` with `min:true`, otherwise I will get invalid `access_hash`

`access_hash` is the `access_hash` where I want to store `channel/group/user`, although it will be updated.

The `username` is also what I want to store. But channel and group are not exposed

Above my reasons. Although I can modify the entities of `chat::group` and `chat::channel` to be `pub` instead of `pub(crate)`, but this requires the introduction of `tl-type` in my project, so I want to expose them

PR #145, opened 2022-09-27
Implement FILE_MIGRATE handling

guzkodes:
This PR implements handling of ```FILE_MIGRATE``` errors when trying to download a file that lives in a different datacenter:

  * Maintains a ```HashMap<i32, FileDownloader>``` at client level
  * Detects the error, picks the ```Sender``` from the session datacenter, exports the auth and imports into a new ```Sender``` that will be stored in the ```HashMap```
  * Redo the request
  
  
  ~~WIP: Depends on https://github.com/Lonami/grammers/pull/146~~

### DIFF
--- a/lib/grammers-client/src/client/client.rs
+++ b/lib/grammers-client/src/client/client.rs
@@ -9,12 +9,12 @@ use crate::utils::{AsyncMutex, Mutex};
 use grammers_mtproto::{mtp, transport};
 use grammers_mtsender::{Enqueuer, Sender};
 use grammers_session::{ChatHashCache, MessageBox, Session};
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
 use std::fmt;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Instant;
-use tokio::sync::Notify;
+use tokio::sync::{Notify, RwLock};
 
 /// When no locale is found, use this one instead.
 const DEFAULT_LOCALE: &str = "en";
@@ -119,6 +119,14 @@ pub(crate) struct ClientInner {
     pub(crate) updates: Mutex<VecDeque<crate::types::Update>>,
     // Used to avoid locking the entire sender when enqueueing requests.
     pub(crate) request_tx: Mutex<Enqueuer>,
+    // Stores per-datacenter downloader instances
+    pub(crate) downloader_map: RwLock<HashMap<i32, Arc<FileDownloader>>>,
+}
+
+pub(crate) struct FileDownloader {
+    pub(crate) sender: AsyncMutex<Sender<transport::Full, mtp::Encrypted>>,
+    pub(crate) request_tx: Mutex<Enqueuer>,
+    pub(crate) stepping_done: Notify,
 }
 
 /// A client capable of connecting to Telegram and invoking requests.
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -21,6 +21,7 @@ use tokio::{
 
 pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
+pub const FILE_MIGRATE_ERROR: i32 = 303;
 const BIG_FILE_SIZE: usize = 10 * 1024 * 1024;
 const WORKER_COUNT: usize = 4;
 
@@ -84,9 +85,15 @@ impl DownloadIter {
 
         use tl::enums::upload::File;
 
-        // TODO handle FILE_MIGRATE and maybe FILEREF_UPGRADE_NEEDED
+        // TODO handle maybe FILEREF_UPGRADE_NEEDED
+        let mut dc: Option<u32> = None;
         loop {
-            break match self.client.invoke(&self.request).await {
+            let result = match dc.take() {
+                None => self.client.invoke(&self.request).await,
+                Some(dc) => self.client.invoke_in_dc(&self.request, dc as i32).await,
+            };
+
+            break match result {
                 Ok(File::File(f)) => {
                     if f.bytes.len() < self.request.limit as usize {
                         self.done = true;
@@ -101,7 +108,11 @@ impl DownloadIter {
                 Ok(File::CdnRedirect(_)) => {
                     panic!("API returned File::CdnRedirect even though cdn_supported = false");
                 }
-                Err(e) => Err(e),
+                Err(InvocationError::Rpc(err)) if err.code == FILE_MIGRATE_ERROR => {
+                    dc = err.value;
+                    continue;
+                }
+                Err(e) => Err(e.into()),
             };
         }
     }
@@ -214,8 +225,8 @@ impl Client {
             let part_index = part_index.clone();
             let client = self.clone();
             let task = tokio::task::spawn(async move {
-                // TODO: use retry_offset to control FILE_MIGRATE retries
                 let mut retry_offset = None;
+                let mut dc = None;
                 loop {
                     // Calculate file offset
                     let offset: i64 = {
@@ -232,15 +243,17 @@ impl Client {
                         break;
                     }
                     // Fetch from telegram
-                    let res = client
-                        .invoke(&tl::functions::upload::GetFile {
-                            precise: true,
-                            cdn_supported: false,
-                            location: location.clone(),
-                            offset,
-                            limit: MAX_CHUNK_SIZE,
-                        })
-                        .await;
+                    let request = &tl::functions::upload::GetFile {
+                        precise: true,
+                        cdn_supported: false,
+                        location: location.clone(),
+                        offset,
+                        limit: MAX_CHUNK_SIZE,
+                    };
+                    let res = match dc {
+                        None => client.invoke(request).await,
+                        Some(dc) => client.invoke_in_dc(request, dc as i32).await,
+                    };
                     match res {
                         Ok(tl::enums::upload::File::File(file)) => {
                             tx.send((offset as u64, file.bytes)).unwrap();
@@ -251,6 +264,11 @@ impl Client {
                             );
                         }
                         Err(InvocationError::Rpc(err)) => {
+                            if err.code == FILE_MIGRATE_ERROR {
+                                dc = err.value;
+                                retry_offset = Some(offset);
+                                continue;
+                            }
                             return Err(InvocationError::Rpc(err));
                         }
                         Err(e) => return Err(e),
--- a/lib/grammers-client/src/client/files.rs
+++ b/lib/grammers-client/src/client/files.rs
@@ -21,7 +21,7 @@ use tokio::{
 
 pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
-pub const FILE_MIGRATE_ERROR: i32 = 303;
+const FILE_MIGRATE_ERROR: i32 = 303;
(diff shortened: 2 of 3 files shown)


guzkodes:
The CI error is in a file I din't change, so likely unrelated to the changes

lib/grammers-client/examples/downloader.rs
```diff
@@ -0,0 +1,195 @@
...
+//! cargo run --example downloader -- CHAT_NAME
+//! ```
+//!
+//! Messages will be printed to stdout, and media will be saved in the ```target/``` folder locally, named
+//! message-[id].[jpg/mp4]
+//!
+extern crate core;
+extern crate grammers_client;
+extern crate grammers_session;
+extern crate grammers_tl_types;
```
Lonami, line 19: Is there any reason to do this? None of the other crate examples do.

lib/grammers-client/examples/downloader.rs
```diff
@@ -0,0 +1,195 @@
...
+    let mut client_handle = client.clone();
+    let network_handle = task::spawn(async move { client.run_until_disconnected().await });
+
+    let chat = client_handle.resolve_username(chat_name.as_str()).await?;
+
+    if chat.is_none() {
+        panic!("Chat {} could not be found", chat_name);
+    }
+
+    let mut messages = client_handle.iter_messages(&chat.unwrap());
```
Lonami, line 112: `is_none` followed by `unwrap` is not very idiomatic. We could change it with `.expect("Chat could not be found")` or maybe a `match`.

lib/grammers-client/examples/downloader.rs
```diff
@@ -0,0 +1,195 @@
...
+    return format!(".{}", suffix);
+}
+
+fn get_subtype(mime_type: Option<&str>) -> String {
+    return mime_type
+        .map(|m| {
+            let mime: Mime = m.parse().unwrap();
+            mime.subtype().to_string()
+        })
+        .unwrap_or(String::new());
```
Lonami, line 180: This might create files such as `message-1001.` (empty extension). Perhaps it would be better to use the extension from the media's file name?

lib/grammers-client/src/client/client.rs
```diff
@@ -119,6 +119,14 @@ pub(crate) struct ClientInner {
     pub(crate) updates: Mutex<VecDeque<crate::types::Update>>,
     // Used to avoid locking the entire sender when enqueueing requests.
     pub(crate) request_tx: Mutex<Enqueuer>,
+    // Stores per-datacenter FileDownloader instances
+    pub(crate) downloader_map: RwLock<HashMap<i32, Arc<FileDownloader>>>,
+}
+
+pub(crate) struct FileDownloader {
```
Lonami, line 126: Not entirely sure if we should name these `FileDownloader`. They're more like generic exported senders after all (which are used for downloads, yes).

lib/grammers-client/src/client/files.rs
```diff
@@ -264,14 +264,10 @@ impl Client {
                             );
                         }
                         Err(InvocationError::Rpc(err)) => {
-                            // Retry on rate limit
-                            if err.code == 420 {
```
Lonami, line 268: Was it intended to remove this retry logic in this PR?

lib/grammers-client/src/client/net.rs
```diff
@@ -263,20 +284,85 @@ impl Client {
...
+                    &new_downloader.sender,
+                    &new_downloader.stepping_done,
+                )
+                .await?;
+
+                mutex.insert(dc_id, new_downloader.clone());
+                Ok(new_downloader.clone())
+            }
+            Err(e) => {
+                panic!("Cannot obtain new sender to datacenter {}", e)
```
Lonami, line 332: Do we really want a panic here, or should this return the error?

lib/grammers-client/src/types/media.rs
```diff
@@ -49,6 +49,94 @@ pub enum Media {
     Contact(Contact),
 }
 
+pub trait MultiSized {
```
Lonami, line 52: I'm okay with improving the way photo sizes but I think this should be in a separate PR (perhaps merged first).

Lonami (commented):
Thanks for the PR! Left some comments. My main concern is with some changes that I would prefer for them to be in a separate PR but I'm not sure if there's a reason why they're in this one (single it's also a single commit, it's harder to justify), along with a questionable public API surface.

lib/grammers-client/examples/downloader.rs
guzkodes: no reason. It is a leftover

lib/grammers-client/examples/downloader.rs
```diff
@@ -0,0 +1,195 @@
...
+                    );
+                    let path = Path::new(dest.as_str());
+                    client_handle
+                        .download_media(&media, &path)
+                        .await
+                        .expect("Error downloading message");
+                }
+            }
+            Err(e) => panic!("{}", e),
+            Ok(None) => break { println!("Finished iterating messages") },
```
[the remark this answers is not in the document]
guzkodes: It does work, but I did it initially like this for debugging purposes, I will refactor to make it more easy on the eyes

lib/grammers-client/examples/downloader.rs
guzkodes: wdym, it will always try to use the mime type if it's present, otherwise it is not possible to know the extension unless inspecting the bytes of the content

lib/grammers-client/src/client/auth.rs
```diff
@@ -180,6 +181,24 @@ impl Client {
         }
     }
 
+    /// Returns data for copying authorization to another data-center.
+    ///
+    /// See https://core.telegram.org/method/auth.exportAuthorization
+    pub async fn export_authorization(
```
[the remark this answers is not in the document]
guzkodes: Initially it was ```pub(crate)``` , I made it ```pub``` because it is part of the telegram API schema (https://core.telegram.org/method/auth.exportAuthorization).
Since the import is non trivial and depends on lots of internal private parts, you're right, I see no reason to expose this.
I will hide this inside ```net.rs```

lib/grammers-client/src/client/client.rs
guzkodes: what about ```DatacenterConnector``` ?

lib/grammers-client/src/client/files.rs
guzkodes: yes, the wait login is inside https://github.com/Lonami/grammers/blob/master/lib/grammers-client/src/client/net.rs#L251-L267

lib/grammers-client/src/client/net.rs
guzkodes: No reason for panic 👍

lib/grammers-client/src/types/media.rs
guzkodes: no problem. I put it all together because without it most downloads fail

lib/grammers-client/src/client/net.rs
guzkodes: How should this ```AuthorizationError``` should be propagated ? it is not compatible with ```InvocationError```

guzkodes: I am thinking in adding a new variant to ```InvocationError``` to signal for security issues

lib/grammers-client/src/types/media.rs
guzkodes: https://github.com/Lonami/grammers/pull/146

lib/grammers-client/examples/downloader.rs
Lonami: I meant there would be a trailing dot (`message-1001.` rather than `message-1001`).

lib/grammers-client/src/client/files.rs
Lonami: Yes but this change is still not directly related to handling `FILE_MIGRATE` which is what I meant (if this change was for this or another PR).

lib/grammers-client/src/client/client.rs
Lonami: Hmm not really... Perhaps we can leave it for now.

lib/grammers-client/src/client/net.rs
Lonami: The method should return `AuthorizationError` instead (because `InvocationError` is compatible with it, just not the other way around).

lib/grammers-client/examples/downloader.rs
guzkodes: Fixed in https://github.com/Lonami/grammers/pull/148

lib/grammers-client/examples/downloader.rs
```diff
@@ -0,0 +1,195 @@
...
+//! [Windows](https://ss64.com/nt/set.html) or [Linux](https://ss64.com/bash/export.html))
+//! to Telegram's API ID and API hash respectively.
+//!
+//! Then, run it as:
+//!
+//! ```sh
+//! cargo run --example downloader -- CHAT_NAME
+//! ```
+//!
+//! Messages will be printed to stdout, and media will be saved in the ```target/``` folder locally, named
```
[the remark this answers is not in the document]
guzkodes: This was already solved in a separate PR

lib/grammers-client/examples/downloader.rs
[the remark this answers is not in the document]
guzkodes: Already solved on separated  PR

lib/grammers-client/src/client/files.rs
guzkodes: I separated it into another PR: https://github.com/Lonami/grammers/pull/149

lib/grammers-client/src/client/net.rs
guzkodes: My question was more like in terms of API breakage,  as this is method is part a chain of methods:
download_media (file.rs)-> next (file.rs) -> invoke_in_dc (net.rs) -> connect_sender (net.rs)
If I change it to return ```AuthorizationError``` the callers will also have to return it and so on

lib/grammers-client/src/client/files.rs
```diff
@@ -21,6 +21,7 @@ use tokio::{
 
 pub const MIN_CHUNK_SIZE: i32 = 4 * 1024;
 pub const MAX_CHUNK_SIZE: i32 = 512 * 1024;
+pub const FILE_MIGRATE_ERROR: i32 = 303;
```
Lonami, line 24: The chunk size is public to let the users know that's the limits they can configure but this error doesn't really need to be exposed to the users:

```suggestion
const FILE_MIGRATE_ERROR: i32 = 303;
```

lib/grammers-client/src/client/net.rs
```diff
@@ -278,6 +299,84 @@ impl Client {
...
+        &self,
+        dc_id: i32,
+    ) -> Result<Option<Arc<FileDownloader>>, InvocationError> {
+        return Ok({
+            let guard = self.0.downloader_map.read().await;
+            guard.get(&dc_id).map(|f| f.clone())
+        });
+    }
+
+    pub async fn invoke_in_dc<R: tl::RemoteCall>(
```
Lonami, line 361: I think it's fine to leave this as public (there's other requests which must be executed in a different datacenter, like getting graph data from channels I think).

lib/grammers-client/src/client/net.rs
guzkodes: marked as resolved, so keep panic?

Lonami:
Thanks! If you ever feel like you need to change something regarding this don't hesitate to send another pull request.

PR #147, opened 2022-09-28
Fix 'API is returning peers without access_hash' assertion error

guzkodes:
Error happens when executing  ```cargo run --example dialogs``` or  ```cargo run --example downloader -- CHAT``` due to a users with ```min: true``` being returned

### DIFF
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -117,10 +117,8 @@ impl ParticipantIter {
                 };
 
                 let mut chat_hashes = client.0.chat_hashes.lock("iter_participants");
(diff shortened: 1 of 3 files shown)


Lonami (commented):
Did you actually check if it can happen in other places except for `get_dialogs`, or did you simply assume it can happen?

guzkodes:
It also happens for messages.rs (see above) and regarding chat.rs I assumed it can happen too since it deals with users from a chat. I can try to reproduce tomorrow

Lonami:
Alright, if it happens with messages (I missed that) and dialogs it's probably safe to assume it happens elsewhere. Thanks!

PR #151, opened 2022-10-11
add join_chat/accept_invite_link method

tingfeng-key:
add join_chat/accept_invite_link method

### DIFF
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -750,8 +750,10 @@ impl Client {
         Ok(None)
     }
 
-    /// Join a group or channel.
-    /// use PackedChat
+    /// Join a public group or channel.
+    ///
+    /// A channel is public if it has a username.
+    /// To join private chats, [`Client::accept_invite_link`](Client::accept_invite_link) should be used instead.
     pub async fn join_chat(
         &mut self,
         packed_chat: PackedChat,
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -725,7 +725,7 @@ impl Client {
         let permissions = ParticipantPermissions::Channel(participant.participant);
         Ok(permissions)
     }
-    
+
     #[cfg(feature = "parse_invite_link")]
     fn parse_invite_link(invite_link: &str) -> Option<String> {
         let url_parse_result = url::Url::parse(invite_link);
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -726,6 +726,7 @@ impl Client {
         Ok(permissions)
     }
     
+    #[cfg(feature = "parse_invite_link")]
     fn parse_invite_link(invite_link: &str) -> Option<String> {
         let url_parse_result = url::Url::parse(invite_link);
         if url_parse_result.is_err() {
@@ -772,7 +773,7 @@ impl Client {
 
         None
     }
-    
+
     /// Accept an invite link to join the corresponding private chat.
     ///
     /// If the chat is public (has a public username), [`Client::join_chat`](Client::join_chat) should be used instead.
@@ -780,21 +781,25 @@ impl Client {
         &mut self,
         invite_link: &str,
     ) -> Result<tl::enums::Updates, InvocationError> {
-        use grammers_mtproto::mtp::RpcError;
-        let parse_result = Self::parse_invite_link(invite_link);
-        if parse_result.is_none() {
-            return Err(InvocationError::Rpc(RpcError {
-                code: 400,
-                name: "INVITE_HASH_INVALID".to_string(),
-                value: None,
-                caused_by: None,
-            }));
-        }
+        #[cfg(not(feature = "parse_invite_link"))]
+        let hash = invite_link.to_string();
+        #[cfg(feature = "parse_invite_link")]
+        let hash = {
+            use grammers_mtproto::mtp::RpcError;
+            let parse_result = Self::parse_invite_link(invite_link);
+            if parse_result.is_none() {
+                return Err(InvocationError::Rpc(RpcError {
+                    code: 400,
+                    name: "INVITE_HASH_INVALID".to_string(),
+                    value: None,
+                    caused_by: None,
+                }));
+            }
+            parse_result.unwrap()
+        };
 
-        self.invoke(&tl::functions::messages::ImportChatInvite {
-            hash: parse_result.unwrap(),
-        })
-        .await
+        self.invoke(&tl::functions::messages::ImportChatInvite { hash })
+            .await
     }
 
     /// Join a public group or channel.
--- a/lib/grammers-client/src/client/chats.rs
+++ b/lib/grammers-client/src/client/chats.rs
@@ -756,27 +756,42 @@ impl Client {
     ///
     /// A channel is public if it has a username.
     /// To join private chats, [`Client::accept_invite_link`](Client::accept_invite_link) should be used instead.
-    pub async fn join_chat(
+    pub async fn join_chat<C: Into<PackedChat>>(
         &mut self,
-        packed_chat: PackedChat,
+        packed_chat: C,
     ) -> Result<Option<Chat>, InvocationError> {
         use tl::enums::Updates;
 
+        let chat = packed_chat.into();
         let update_chat = match self
             .invoke(&tl::functions::channels::JoinChannel {
-                channel: packed_chat.try_to_input_channel().unwrap(),
+                channel: chat.try_to_input_channel().unwrap(),
             })
             .await?
         {
-            Updates::Combined(updates) => updates.chats.first().cloned(),
-            Updates::Updates(updates) => updates.chats.first().cloned(),
+            Updates::Combined(updates) => Some(
+                updates
+                    .chats
+                    .into_iter()
+                    .filter(|x| x.id() == chat.id)
+                    .collect::<Vec<tl::enums::Chat>>(),
+            ),
+            Updates::Updates(updates) => Some(
+                updates
+                    .chats
+                    .into_iter()
+                    .filter(|x| x.id() == chat.id)
+                    .collect::<Vec<tl::enums::Chat>>(),
+            ),
             _ => None,
         };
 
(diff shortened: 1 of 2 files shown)


lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
         let permissions = ParticipantPermissions::Channel(participant.participant);
         Ok(permissions)
     }
+    
+    /// join private chat
```
Lonami, line 729:
```suggestion
    /// Accept an invite link to join the corresponding private chat.
    ///
    /// If the chat is public (has a public username), [`Client::join_chat`](Client::join_chat) should be used instead.
```

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
...
+    pub async fn accept_invite_link(
+        &mut self,
+        invite_link: &str,
+    ) -> Result<Option<Chat>, InvocationError> {
+        use tl::enums::Updates;
+        assert!(invite_link.starts_with("https://t.me/joinchat/"));
```
Lonami, line 735: We can't have asserts like this. The code should return an error if the link is not valid.

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
...
+        let update_chat = match self
+            .invoke(&tl::functions::messages::ImportChatInvite {
+                hash: invite_link.replace("https://t.me/joinchat/", ""),
```
Lonami, line 738: This is very brittle. There is a lot more invite links Telegram supports (for example, the `https` could be `http`, `t.me` could be `telegram.me` or even `telegram.dog`, and `joinchat/` could be just `+` among others).

This code should properly validate the invite link and extract the join hash, rather than simply replacing it away.

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
...
+            })
+            .await?
+        {
+            Updates::Combined(updates) => updates.chats.first().cloned(),
+            Updates::Updates(updates) => updates.chats.first().cloned(),
```
Lonami, line 743: Is it true that the joined chat will always be the first one?

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
...
+            _ => None,
+        };
+
+        if let Some(chat) = update_chat {
+            return Ok(Some(Chat::from_chat(chat)));
+        }
+        Ok(None)
```
Lonami, line 750: Can this actually happen?

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,55 @@ impl Client {
...
+    }
+
+    /// Join a group or channel.
+    /// use PackedChat
+    pub async fn join_chat(
+        &mut self,
+        packed_chat: PackedChat,
```
Lonami, line 757: This should work like `send_message`, accepting anything which can be a chat. Not only with the exact `PackedChat` type.

Lonami (changes requested):

lib/grammers-client/src/client/chats.rs
tingfeng-key: Here I can think of the way to deal with it is to introduce regex, is there any other better way?

lib/grammers-client/src/client/chats.rs
tingfeng-key: It's ok, is there any problem here?

lib/grammers-client/src/client/chats.rs
tingfeng-key: To return an error, define a new error type, like auth::SignInError, is there any other way?

Lonami: I think we should create a fake RPC error like this:

https://github.com/Lonami/grammers/blob/c159fb086224342a4a45f719874cc675a0db5e46/lib/grammers-client/src/types/chats.rs#L69-L74

but with `name: "INVITE_HASH_INVALID"`.

lib/grammers-client/src/client/chats.rs
Lonami: If this can never occur, I would prefer `update_chat` being `None` to `panic!()`. This way the method can always return the chat and not just an `Option`.

If this can occur, we should document when it can happen.

lib/grammers-client/src/client/chats.rs
Lonami: We have two options:

* Let the user deal with it. The user should provide the hash, not a link.
* Parse the URL. We can write a "parser" for the few supported URLs, or add a new dependency on a crate capable of parsing URLs. Once the URL is parsed, validate the domain and extract the hash from the path.

I would prefer to avoid depending on a regex crate if possible.

tingfeng-key: Well, when I use option two to modify (the `url` crate used), I almost crash. Although I think the checking of the rules is done, when I want to get the hash in the URL, I find that a lot of replacements are needed. Maybe my way is too simplistic
```rust
use grammers_mtproto::mtp::RpcError;
        let url_parse = url::Url::parse(invite_link).map_err(|_| {
            InvocationError::Rpc(RpcError {
                code: 400,
                name: "INVITE_HASH_INVALID".to_string(),
                value: None,
                caused_by: None,
            })
        })?;

        let (scheme, host, path) = (url_parse.scheme(), url_parse.host_str(), url_parse.path());

        if !vec!["https", "http"].contains(&scheme)
            || host.is_none()
            || (!path.starts_with("joinchat/") || !path.starts_with("+"))
        {
            return Err(InvocationError::Rpc(RpcError {
                code: 400,
                name: "INVITE_HASH_INVALID".to_string(),
                value: None,
                caused_by: None,
            }));
        }

        let host = host.unwrap();
        let hosts = host.split(".").collect::<Vec<&str>>();
        let host_check_result = match hosts.len() {
            2 if vec!["t", "telegram"].contains(&hosts[0])
                && vec!["me", "org", "dog"].contains(&hosts[0]) =>
            {
                true
            }
            3 if "www" == hosts[0]
                && vec!["t", "telegram"].contains(&hosts[1])
                && vec!["me", "org", "dog"].contains(&hosts[2]) =>
            {
                true
            }
            _ => false,
        };

        if !host_check_result {
            return Err(InvocationError::Rpc(RpcError {
                code: 400,
                name: "INVITE_HASH_INVALID".to_string(),
                value: None,
                caused_by: None,
            }));
        }
```

tingfeng-key: Both option 1 and option 2 are supported

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,124 @@ impl Client {
...
+        }
+
+        let url_parse = url_parse_result.unwrap();
+        let scheme = url_parse.scheme();
+        let path = url_parse.path();
+        if url_parse.host_str().is_none() || !vec!["https", "http"].contains(&scheme) {
+            return None;
+        }
+        let host = url_parse.host_str().unwrap();
+        let hosts = vec![
```
Lonami, line 743: This does not need to be a `vec![]`. A simple array `[]` should work too.

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,124 @@ impl Client {
...
+    ///
+    /// If the chat is public (has a public username), [`Client::join_chat`](Client::join_chat) should be used instead.
+    pub async fn accept_invite_link(
+        &mut self,
+        invite_link: &str,
+    ) -> Result<tl::enums::Updates, InvocationError> {
+        #[cfg(not(feature = "parse_invite_link"))]
+        let hash = invite_link.to_string();
+        #[cfg(feature = "parse_invite_link")]
+        let hash = {
```
Lonami, line 787: This is rather confusing. A crate feature should generally not change the expected format of input parameters.

Imagine I depend on crate `foo` which depends on `grammers-client`. `foo` doesn't have the feature enabled, so it uses the hash. Now, I also depend on `grammers-client`, but enable the feature. Now when I use `foo`, it won't work correctly, because the feature is enabled.

I think we should get rid of the feature and make the `url` dependency mandatory (so it will always be available).

lib/grammers-client/src/client/chats.rs
```diff
@@ -725,6 +725,124 @@ impl Client {
...
+            let parse_result = Self::parse_invite_link(invite_link);
+            if parse_result.is_none() {
+                return Err(InvocationError::Rpc(RpcError {
+                    code: 400,
+                    name: "INVITE_HASH_INVALID".to_string(),
+                    value: None,
+                    caused_by: None,
+                }));
+            }
+            parse_result.unwrap()
```
Lonami, line 798: This should use a `match` or at least use `Option::ok_or`.

Checking for `none` and then using `unwrap()` is not idiomatic (clippy might lint against this, not sure).

Lonami (changes requested):

lib/grammers-client/src/client/chats.rs
Lonami: To clarify what I meant in https://github.com/Lonami/grammers/pull/151#discussion_r992472784: we should pick one option only. I think it's fine to always have `url` available and accept all forms of invite links (option 2 only).

tingfeng-key: So do I need to modify parse_invite_link to become the url?

tingfeng-key: I removed the parameter to pass the hash directly, and updated it to accept_invite_link only when feature = "parse_invite_link" is enabled. Is this normal?

Lonami: Not what I had in mind but that'll work. We can change it later.

Lonami:
Bump. CI failed because it did not pass the `rustfmt` check. Once it's green I can review again.

Lonami:
You should be able to run `cargo fmt` to fix the CI error.

tingfeng-key:
But when I execute cargo run in my patch-2 branch, it doesn't give any hint and no output

PR #152, opened 2022-10-11
add to_input_channel_lossy method

tingfeng-key:
add to_input_channel_lossy method

### DIFF
--- a/lib/grammers-session/src/chat/packed.rs
+++ b/lib/grammers-session/src/chat/packed.rs
@@ -170,6 +170,16 @@ impl PackedChat {
             _ => None,
         }
     }
+    
+    pub fn to_input_channel_lossy(&self) -> tl::enums::InputChannel {
+        self.try_to_input_channel().unwrap_or_else(|| {
+            tl::types::InputChannel {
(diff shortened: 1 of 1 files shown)


Lonami:
What is the need for this?

tingfeng-key:
Unwrap is required when I call try_to_input_channel way.
I want to avoid calling unwrap outside (just for this reason).
When I looked at the source code, I found to_input_user_lossy, so I added this method 😬

Lonami:
`to_input_user_lossy` is used by the library, but `to_input_channel_lossy` is not.

To be fair, `to_input_user_lossy` itself should probably be cleaned up. I do not think it's a good idea to add more methods like this to `PackedChat`. They're more of a workaround than an actual solution.

You can keep using the method in your own code if you need the conversion:

```python
    pub fn to_input_channel_lossy(chat: &PackedChat) -> tl::enums::InputChannel {
        chat.try_to_input_channel().unwrap_or_else(|| {
            tl::types::InputChannel {
                channel_id: 0,
                access_hash: 0,
            }
            .into()
        })
    }
```

PR #153, opened 2022-10-11
add invite_to_channel/add_chat_user method

tingfeng-key:
add invite_to_channel/add_chat_user method


Lonami:
I'm a bit wary of adding these methods to the library. Adding users without their consent is considered spam, so I don't know if we should make it any easier. I can't stop people from calling these methods (and stopping them entirely would be dumb), but I can at least make it a bit more annoying to do so.

Telethon also doesn't offer these methods as client methods. For the time being, I think raw API should keep being used to add users, if you really need to.

tingfeng-key:
OK, I get it. At the moment, I need it myself, and I have done the implementation of tl conversion, so I submit it to pr

PR #157, opened 2022-10-23
Refactor code with rust clippy

saguywalker:


lib/grammers-client/src/client/updates.rs
```diff
@@ -56,8 +56,8 @@ impl Client {
                 let mut message_box = self.0.message_box.lock("client.next_update");
                 // This temporary is needed or message_box's lifetime is extended too much.
                 // See https://github.com/rust-lang/rust/issues/102423.
-                let diff = message_box.get_difference();
-                diff
+                
+                message_box.get_difference()
```
Lonami, line 60: This removes the temporary which I think will affect the above comment.

Lonami (commented):
Other than the temporary variable change, looks good.

Lonami:
CI complains about having to run `cargo fmt` on the project. Once that's done I'll rerun it to check for the temporary variable.

PR #161, opened 2022-12-21
Add more filters to search messages

guzkodes:


lib/grammers-client/src/client/messages.rs
```diff
@@ -262,6 +264,48 @@ impl SearchIter {
...
+    /// // Search messages only from Dec/2022
+    /// let start_date = DateTime::parse_from_rfc3339("2021-12-01T00:00:00-00:00").unwrap();
+    /// let end_date = DateTime::parse_from_rfc3339("2021-12-31T23:59:59-00:00").unwrap();
+    ///
+    /// let mut messages = client.search_messages(&chat).date_interval(&Some(start_date), &Some(end_date));
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn date_interval(
```
Lonami, line 295: What's your opinion on offering two methods for this instead of one?

guzkodes: Usually we want to conditionally fill either one or both dates, and since the filter methods take a `self` reference, it would consume the SearchIter reference after one call and throw "cannot move" error afterwards.

Lonami: Sorry, I'm not sure I fully understand. The functions take `self` by value and return `Self` back, so they can be chained (`let iter = client.search_messages(chat).query("foo").filter(type);`). What code causes the problem you describe?

guzkodes: Something like this causes problem:

```
let iter = client.search_messages(chat):
if !min_date.is_empty() {
     iter.from_date(min_date);
}
if !max_date.is_empty() {
     iter.to_date(max_date);
}
```

guzkodes: But that was before I added ```Option``` granted, probably it can split into 2 methods, I don't have any preference

Lonami: Yeah the builder pattern in Rust is full of tradeoffs... in your case you should be able to do this:

```rust
let mut iter = client.search_messages(chat);
if !min_date.is_empty() {
     iter = iter.from_date(min_date.unwrap());
}
if !max_date.is_empty() {
     iter = iter.to_date(max_date.unwrap());
}
```

So, yes, I would prefer separate methods. Not sure what to call them though. `from` or `min`. `to` or `max`. `oldest` or `newest`? No idea. I'm open for suggestions. Or we can just pick whichever and it'll be fine anyway.

guzkodes: oh I see, I missed the reassign part. Updated PR to use two methods

lib/grammers-client/src/client/messages.rs
```diff
@@ -262,6 +264,50 @@ impl SearchIter {
         self
     }
 
+    /// Restricts results to messages sent by the logged-in user
+    pub fn sent_by_self(mut self) -> Self {
```
Lonami, line 268: Would probably make sense to accept anything that can be converted into a `PackedChat` instead (like the rest of methods do).

PR #171, opened 2023-04-07
Fix up clippy lints

Architector4:
Here be 17 commits that manually fix a bunch of lints highlighted by Cargo's Clippy.

None should alter the code's behavior and should amount to make the code a bit cleaner.


Lonami:
Thanks!

PR #172, opened 2023-04-07
Account for less than 2 salts in request buffer

Architector4:
This change accounts for a case when there is only 0 or 1 salt in the request buffer, instead of panicking on a subtraction underflow error.

The panic happened to me regularly when Grammers is connected to a Telegram test server regularly, and I hope this should fix that from happening.

### DIFF
--- a/lib/grammers-mtproto/src/mtp/encrypted.rs
+++ b/lib/grammers-mtproto/src/mtp/encrypted.rs
@@ -1156,11 +1156,16 @@ impl Mtp for Encrypted {
         // Check to see if the next salt can be used already. If it can, drop the current one and,
         // if the next salt is the last one, fetch more.
         if let Some((start_secs, start_instant)) = self.start_salt_time {
-            if let Some(salt) = self.salts.get(self.salts.len() - 2) {
+            if let Some(salt) = self
+                .salts
+                .len()
+                .checked_sub(2)
+                .and_then(|x| self.salts.get(x))
+            {
                 let now = start_secs + start_instant.elapsed().as_secs() as i32;
                 if now >= salt.valid_since + SALT_USE_DELAY {
                     self.salts.pop();
-                    if self.salts.len() == 1 {
+                    if self.salts.len() <= 1 {
                         info!("only one future salt remaining; asking for more salts");
                         let body = tl::functions::GetFutureSalts {
                             num: NUM_FUTURE_SALTS,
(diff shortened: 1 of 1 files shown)


lib/grammers-mtproto/src/mtp/encrypted.rs
```diff
@@ -1156,11 +1156,11 @@ impl Mtp for Encrypted {
         // Check to see if the next salt can be used already. If it can, drop the current one and,
         // if the next salt is the last one, fetch more.
         if let Some((start_secs, start_instant)) = self.start_salt_time {
-            if let Some(salt) = self.salts.get(self.salts.len() - 2) {
+            if let Some(salt) = self.salts.len().checked_sub(2).map(|x| &self.salts[x]) {
                 let now = start_secs + start_instant.elapsed().as_secs() as i32;
                 if now >= salt.valid_since + SALT_USE_DELAY {
                     self.salts.pop();
-                    if self.salts.len() == 1 {
+                    if self.salts.len() <= 1 {
```
Lonami, line 1163: `salts` will never be 0 here or the `checked_sub` would've been `None`.

Lonami (changes requested):
Perhaps what we should do is check "if there is only one salt left, and it's about to expire, fetch the next batch before it does". Since `future_salt` is defined as:

```
future_salt#0949d9dc valid_since:int valid_until:int salt:long = FutureSalt;
```

we should be able to use `valid_until` instead.

lib/grammers-mtproto/src/mtp/encrypted.rs
Architector4: > `salts` will never be 0 here or the `checked_sub` would've been `None`.

fiiiiine changed it back lol

Architector4:
> Perhaps what we should do is check "if there is only one salt left, and it's about to expire, fetch the next batch before it does".

That makes sense; you did mention in the Telegram chat that this section of code is due a rewrite in general, and that you might get around to it someday.

However, my goal here was to implement a quick fix for an issue I was experiencing (the library panicking when such a scenario occurs), and other than that I'm not familiar enough with the logic and semantics of these salts or how to manage them to provide anything more adequate than that.

I believe merging this would still be helpful as it prevents the aforementioned crash until the rewrite, but your call of course. Maybe the change I made makes no sense and actually breaks everything for all I know, to be honest :v

Lonami:
Closing this in favor of #173.

PR #176, opened 2023-04-11
Add some getters and unstable_raw feature

akshettrj:
- `as_raw` for Group, Channel and User under "unstable_raw"
- `is_animated` and `is_spoiler` for Document media type
- `is_spoiler` for Photo media type
- `access_hash` and `file_reference` for Photo and Document media types under "unstable_raw"

### DIFF
--- a/lib/grammers-client/src/types/media.rs
+++ b/lib/grammers-client/src/types/media.rs
@@ -150,6 +150,11 @@ impl Photo {
                 .collect(),
         }
     }
+
+    /// Returns true if the photo is a spoiler.
+    pub fn is_spoiler(&self) -> bool {
+        self.photo.spoiler
+    }
 }
 
 impl Document {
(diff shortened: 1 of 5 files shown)


lib/grammers-client/src/types/media.rs
```diff
@@ -150,6 +150,31 @@ impl Photo {
...
         }
     }
+
+    /// Returns true if the photo is a spoiler.
+    pub fn is_spoiler(&self) -> bool {
+        self.photo.spoiler
+    }
+
+    #[cfg(feature = "unstable_raw")]
+    pub fn access_hash(&self) -> Option<i64> {
```
Lonami, line 160: `unstable_raw` should only add `as_raw` methods.

Lonami (changes requested):
Thanks for the PR, but I would prefer not to merge as-is.

Lonami:
Thanks!

PR #177, opened 2023-04-14
Add some missing media types

akshettrj:
I have added some unimplemented media types and added some getters for them.

Please let me know if you want them to be implemented in a different way


Lonami (commented):
Seems a bit annoying that `Geo` getters have to return `Option` all the time. Perhaps it would be better to treat `GeoEmpty` as "no media", so the getters wouldn't need to worry about it. WDYT?

PR #178, opened 2023-04-14
Add ability to convert Media enum into raw MessageMedia enum

akshettrj:
Added `From<Media>` implementation for `tl::enums::MessageMedia` under `unstable_raw` feature


Lonami (commented):
My question with this then is, what happens with the `as_raw` we added before? It feels inconsistent to keep both (or only one depending on the scenario).

Lonami:
Should we then get rid of `as_raw`'s and add `From` instead?

PR #184, opened 2023-04-27
Make PasswordToken cloneable

roj1512:


Lonami (commented):
Why is this change needed?

Paryyysaa (approved):

Lonami:
Good point. I would prefer if another solution was chosen, though, as making cloneable defeats the purpose of "making invalid states irrepresentable" (here, being able to use the token freely, even after successful sign-in).

What we should do is change the return type of `check_password` so that it includes the token every time. Then you can reuse the token if and only if the method failed and you're not logged in.

But I know that's a pain.

So please at least add:

```rs
// TODO this should not be Clone, but check_password Err doesn't include it back yet
```

above the change so we can remember to address it in the future. And then we can merge this temporary workaround.

(Unless you want to implement the better `Err`. That's the better solution.)

roj1512:
> `check_password Err doesn't include it back yet`

`check_password` error or `SignInError::PasswordInvalid`?

Lonami:
The `Error` type used in `check_password`'s `Err` variant should probably be a new `enum` specific to that method.

We probably want something as discussed in https://t.me/gramme_rs/17454. But that can come later. I'm happy with adding the TODO for now.

roj1512:
Done!

PR #187, opened 2023-05-24
add `is_admin` helper method to tl::types::Channel

MrAliSalehi:
add a method to check if the `current user` is the owner of channel or has admin rights

### DIFF
--- a/lib/grammers-client/src/types/chat/channel.rs
+++ b/lib/grammers-client/src/types/chat/channel.rs
@@ -126,18 +126,25 @@ impl Channel {
         self.0.username.as_deref()
     }
 
-    /// check if the current user has admin rights or ownership in the channel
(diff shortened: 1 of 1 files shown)


lib/grammers-client/src/types/chat/channel.rs
```diff
@@ -126,18 +126,25 @@ impl Channel {
...
-            true
-        } else {
-            if self.0.admin_rights.is_some() {
-                let tl::enums::ChatAdminRights::Rights(rights) =
-                    self.0.admin_rights.as_ref().unwrap();
-                rights.post_messages || rights.change_info || rights.other
-            } else {
-                false
-            }
+    /// return the permissions of the user in the Channel
```
Lonami, line 129:
```suggestion
    /// Return the permissions of the logged-in user in this channel.
```

Issue #150, opened 2022-10-11
HISTORY_GET_FAILED caused by updates.getChannelDifference


tingfeng-key:
When I run it for a period of time (the last time was 10 minutes), it has an error:
```rust
Error: request error: rpc error 500: HISTORY_GET_FAILED caused by updates.getChannelDifference
```

Lonami:
The library needs to handle all 500 errors in this `match`:

https://github.com/Lonami/grammers/blob/96f61a16b4c4ba5e33175fbaafae061dcbd4af25/lib/grammers-client/src/client/updates.rs#L80-L82

The handling is essentially the same as that of `PERSISTENT_TIMESTAMP_OUTDATED`, so we can simply update the condition to enter that arm when either this or the error 500 occurs.

You can refer to [Telethon's implementation](https://github.com/LonamiWebs/Telethon/blob/db29e9b7efb204fbe56fe8d59ccc4d55ceddc165/telethon/client/updates.py#L308) as well if you need more context.

Please feel welcome to send a pull request (it should be simple, just enter the `Err` branch if the error is 500 too).

PR #190, opened 2023-07-08
RPC 500 error handling

tingfeng-key:
RPC 500 error handling #150

### DIFF
--- a/lib/grammers-client/src/client/updates.rs
+++ b/lib/grammers-client/src/client/updates.rs
@@ -124,7 +124,7 @@ impl Client {
                         message_box.end_channel_difference(&request, PrematureEndReason::Banned);
                         continue;
                     }
-                    Err(e) => Err(e) => match e {
+                    Err(e) => match e {
                         InvocationError::Rpc(e) => {
                             if e.code == 500 {
                                 let mut message_box = self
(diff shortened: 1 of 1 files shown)


lib/grammers-client/src/client/updates.rs
```diff
@@ -124,6 +124,18 @@ impl Client {
                         message_box.end_channel_difference(&request, PrematureEndReason::Banned);
                         continue;
                     }
+                    Err(InvocationError::Rpc(rpc_error)) if rpc_error.code.eq(&500) => {
```
Lonami, line 127: Could be just `== 500`.

Lonami (changes requested):

tingfeng-key:
I want to add a log to him, but I don't know what is more appropriate to output

Lonami:
You can probably log a warning, since this error should hopefully be rare enough.

tingfeng-key:
I have added it, you can see if you need to modify it

lib/grammers-client/src/client/updates.rs
```diff
@@ -125,6 +125,7 @@ impl Client {
                         continue;
                     }
                     Err(InvocationError::Rpc(rpc_error)) if rpc_error.code == 500 => {
+                        log::warn!("rpc 500 err: {:#?}", rpc_error);
```
Lonami, line 128:
```suggestion
                        log::warn!("Telegram is having internal issues: {:#?}", rpc_error);
```

PR #195, opened 2023-09-27
Fix panic on IO errors in connect_sender

ByteVanguard:
Replaces `Cannot obtain new sender to datacenter authorization error: request error` panic with error propagation


Lonami:
Thanks.

PR #214, opened 2023-12-26
Extracted dependencies into workspace deps

akorchyn:
It's easier to maintain that way :) 
Hope this small PR will be helpful.


Lonami:
I appreciate the PR, but I'd personally prefer to keep it the way they are. I don't find the current approach "harder to maintain" and I like that they can, for the most part, be separate.

PR #224, opened 2024-04-04
Make document a public field

velaton618:


Lonami (commented):
Public raw API types should be gated behind the unstable-raw feature. This PR also has a lot of unrelated formatting changes. So I will be closing it.

PR #230, opened 2024-04-27
Fixing incorrect password hash calculations

rs-pro0:
Previously you would get "PASSWORD_HASH_INVALID" for correct password sometimes. It should be fixed now.

### DIFF
--- a/lib/grammers-crypto/src/two_factor_auth.rs
+++ b/lib/grammers-crypto/src/two_factor_auth.rs
@@ -7,7 +7,8 @@
 // except according to those terms.
 use glass_pumpkin::safe_prime;
 use hmac::Hmac;
-use num_bigint::BigUint;
+use num_bigint::{BigInt, BigUint, Sign};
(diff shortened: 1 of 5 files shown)


lib/grammers-crypto/src/two_factor_auth.rs
```diff
@@ -64,12 +64,17 @@ pub fn calculate_2fa(
...
         big_g_b - k_v
     } else {
+        changed_sign = true;
         k_v - big_g_b
     };
-    let big_t = sub % &big_p;
+    let mut big_t = sub % &big_p;
+    if changed_sign {
+        big_t = (&big_p - big_t) % &big_p;
+    }
```
Lonami, line 77: We could avoid mut variables by doing the equivalent:

```suggestion
    let sub = big_g_b - k_v;
    let big_t = if sub < BigUint::from(0u32) {
        (sub % &big_p) + &big_p
    } else {
        sub % &big_p
    };
```

PR #231, opened 2024-04-27
Bump to layer 179

rs-pro0:


lib/grammers-session/src/message_box/adaptor.rs
```diff
@@ -505,6 +509,22 @@ impl PtsInfo {
             DeleteQuickReply(_) => None,
             QuickReplyMessage(_) => None,
             DeleteQuickReplyMessages(_) => None,
+            BotBusinessConnect(_) => None,
+            BotNewBusinessMessage(u) => Some(Self {
+                pts: u.qts,
+                pts_count: 0,
```
Lonami, line 515: This *probably* needs to be `1` because there's a new message incoming.

lib/grammers-session/src/message_box/adaptor.rs
```diff
@@ -505,6 +509,22 @@ impl PtsInfo {
...
+                entry: Entry::SecretChats,
+            }),
+            BotEditBusinessMessage(u) => Some(Self {
+                pts: u.qts,
+                pts_count: 0,
```
Lonami, line 520: Similarly, this probably also needs to be 1 because there is 1 message affected.

lib/grammers-session/src/message_box/adaptor.rs
```diff
@@ -505,6 +509,22 @@ impl PtsInfo {
...
+                entry: Entry::SecretChats,
+            }),
+            BotDeleteBusinessMessage(u) => Some(Self {
+                pts: u.qts,
+                pts_count: 0,
```
Lonami, line 525: And here, we likely need `u.messages.len()`.

lib/grammers-tl-types/tl/api.tl
```diff
@@ -1890,9 +1978,9 @@ messages.getChats#49e9528f id:Vector<long> = messages.Chats;
 messages.getFullChat#aeb00b34 chat_id:long = messages.ChatFull;
 messages.editChatTitle#73783ffd chat_id:long title:string = Updates;
 messages.editChatPhoto#35ddd674 chat_id:long photo:InputChatPhoto = Updates;
-messages.addChatUser#f24753e3 chat_id:long user_id:InputUser fwd_limit:int = Updates;
+messages.addChatUser#cbc6d107 chat_id:long user_id:InputUser fwd_limit:int = messages.InvitedUsers;
```
Lonami, line 1981: Damn, another instance of Telegram shoving updates behind things that aren't updates.

We will need to add this new type here:
https://github.com/Lonami/grammers/blob/aa8d52762d5680abbb2deeb6e639416c81d4714c/lib/grammers-mtproto/src/mtp/encrypted.rs#L28-L36

and handle it inside this `extend`:
https://github.com/Lonami/grammers/blob/aa8d52762d5680abbb2deeb6e639416c81d4714c/lib/grammers-mtsender/src/lib.rs#L561

Lonami:
Now looking at Telethon (since that also needs to handle the invited users thing), I've noticed it handles a few more types:

https://github.com/LonamiWebs/Telethon/blob/1974b663a29bbd97ff70bd249f49d899a5cf765e/telethon/network/mtprotosender.py#L693-L697

Those are probably also needed in grammers. Maybe best to leave that correction to a separate PR and just fix `messages.InvitedUsers` here.

lib/grammers-mtproto/src/mtp/encrypted.rs
```diff
@@ -25,14 +25,15 @@ const NUM_FUTURE_SALTS: i32 = 64;
...
-static UPDATE_IDS: [u32; 7] = [
+static UPDATE_IDS: [u32; 8] = [
     tl::types::UpdateShortMessage::CONSTRUCTOR_ID,
     tl::types::UpdateShortChatMessage::CONSTRUCTOR_ID,
     tl::types::UpdateShort::CONSTRUCTOR_ID,
     tl::types::UpdatesCombined::CONSTRUCTOR_ID,
     tl::types::Updates::CONSTRUCTOR_ID,
     tl::types::UpdateShortSentMessage::CONSTRUCTOR_ID,
     tl::types::messages::AffectedMessages::CONSTRUCTOR_ID,
+    tl::types::messages::InvitedUsers::CONSTRUCTOR_ID,
```
Lonami, line 36: We still need to handle this new type in https://github.com/Lonami/grammers/blob/aa8d52762d5680abbb2deeb6e639416c81d4714c/lib/grammers-mtsender/src/lib.rs#L561.

PR #233, opened 2024-04-28
examples in readme + export login token

velaton618:


Lonami (commented):
I appreciate the PR but I don't think adding Rust code to the README is a good idea. It will very quickly stop compiling.

The README already links to the examples folder, so I think that's enough.

As for `export_login_token`, I'd prefer if it didn't return a raw type.

So I'll be closing the PR.

PR #236, opened 2024-05-10
Update Mtproto TL types

TrinhTrungDung:


lib/grammers-client/examples/dialogs.rs
```diff
@@ -42,8 +40,8 @@ async fn async_main() -> Result<()> {
         .init()
         .unwrap();
 
-    let api_id = env!("TG_ID").parse().expect("TG_ID invalid");
-    let api_hash = env!("TG_HASH").to_string();
+    let api_id = std::env::var("TG_ID")?.parse().expect("TG_ID invalid");
```
Lonami, line 43: Please revert this. The use of `env!` is intentional, because developers should be compiling the values into the hinary.

lib/grammers-mtproto/src/authentication.rs
Lonami, line 1: The changes in this file don't seem correct.

lib/grammers-tl-types/tl/mtproto.tl
Lonami, line 1: While `bytes` and `string` are synonyms in TL, they're definitely not in Rust.

The Rust String must contain valid UTF-8, and none of the changed types in this file are guaranteed to be valid UTF-8 (and most likely they aren't).

In general, this file shouldn't be updated.

Lonami (changes requested):
`mtproto.tl` doesn't need to be updated (not under normal circumstances anyway).

I'd be fine merging some of the other improvements though, which I suspect were all linted by a tool like Clippy. But we need to revert `mtproto.tl`.

lib/grammers-client/examples/dialogs.rs
hlf20010508: What about add a `.cargo/config.toml`
```toml
[env]
TG_ID = ""
TG_HASH = ""
```
to avoid error message in example files.

Lonami: If that works, and they can still by overriden by the user, sure.

MOZGIII:
There's a bunch of changes mushed together in a single PR, but it is worth separating them in to separate PRs and discussing separately.
I'd recommend just closing this, especially since it's stale.

PR #239, opened 2024-05-27
create grammers session from telethon string session

taimast:


Lonami:
I am not willing to maintain this, so I won't merge it upstream. My suggested answer to simply login again remains.

PR #241, opened 2024-06-06
MessageIter: add min_id

MaxVerevkin:


Lonami (commented):
I discussed this in the gramme_rs Telegram group previously. I do not see a need for this. One can simply `break` the loop when the `min_id` is reached.

Most requests have `offset_id` but not `max_id` so I prefer the consistency (and `min_id` is not needed because the loop can be broken outside).

MaxVerevkin:
Sorry, didn't know this has already been discussed.

PR #243, opened 2024-06-09
reexport other grammers crates from grammers-client

MaxVerevkin:
So that users of the library don't have to add all these dependencies manually, matching all the versions.

### DIFF
--- a/lib/grammers-client/src/lib.rs
+++ b/lib/grammers-client/src/lib.rs
@@ -46,3 +46,9 @@ pub(crate) mod utils;
 
 pub use client::{Client, Config, InitParams, SignInError};
 pub use types::{button, reply_markup, ChatMap, InputMessage, Update};
+
+pub use grammers_mtproto::transport;
+pub use grammers_mtsender::{FixedReconnect, InvocationError, NoReconnect, ReconnectionPolicy};
+pub use grammers_session as session;
+#[cfg(feature = "unstable_raw")]
+pub use grammers_tl_types;


lib/grammers-client/src/lib.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) mod utils;
 
 pub use client::{Client, Config, InitParams, SignInError};
 pub use types::{button, reply_markup, ChatMap, InputMessage, Update};
+
+pub use grammers_crypto;
```
Lonami, line 50: I'm not sure we should re-export this. When would a user need it when using the client?

lib/grammers-client/src/lib.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) mod utils;
...
+pub use grammers_mtproto;
```
Lonami, line 51: I don't think we need to re-export all from `mtproto`. Realistically, one only needs the `::transport`s from it.

lib/grammers-client/src/lib.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) mod utils;
...
+pub use grammers_mtsender;
```
Lonami, line 52: I'd also prefer to re-export only what's needed here (so errors and reconnection policy)

lib/grammers-client/src/lib.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) mod utils;
...
+pub use grammers_session;
```
Lonami, line 53: I think it's fine to re-export all of this, but making the user type `grammers_client::grammers_session` is a bit of a mouthful. There is only one session. So we could rename this to `pub use grammers_session as session`.

lib/grammers-client/src/lib.rs
```diff
@@ -46,3 +46,9 @@ pub(crate) mod utils;
...
+pub use grammers_tl_types;
```
Lonami, line 54: The ultimate goal is for `grammers-client` to never expose or accept raw types as input, so I'm not sure we should always export this either.

Maybe we should gate it behind `#[cfg(feature = "unstable_raw")]`.

Lonami (changes requested):

Lonami:
You'll need to rebase from master before CI is green here as well.

lib/grammers-client/src/lib.rs
MaxVerevkin: I reexported everything, just in case. Personally I only need `session`, `tl-types` and `mtsender`. I'm happy to remove this.

lib/grammers-client/src/lib.rs
MaxVerevkin: So, `pub use grammers_mtproto::transport as mtproto_transport`? Although I think it would be better to just reexport the entire crate. How about `pub mod reexports { ... }`?

Lonami: The underlying transport protocols one can use in the client are the same as those the MTP supports. So I think it'd be fine to simply `pub use grammers_mtproto::transport`.

In user's code, `use grammers_client::transport::Intermediate` reads fine to me.

I don't think the concept of transports is used for anything else in the high-level API.

PR #244, opened 2024-06-09
fix "attempt to subtract with overflow"

MaxVerevkin:
```
thread 'main' panicked at /home/maxv/rust/grammers/lib/grammers-mtproto/src/mtp/encrypted.rs:1205:48:
attempt to subtract with overflow
````

### DIFF
--- a/lib/grammers-mtproto/src/mtp/encrypted.rs
+++ b/lib/grammers-mtproto/src/mtp/encrypted.rs
@@ -1202,7 +1202,8 @@ impl Mtp for Encrypted {
         // Check to see if the next salt can be used already. If it can, drop the current one and,
         // if the next salt is the last one, fetch more.
         if let Some((start_secs, start_instant)) = self.start_salt_time {
-            if let Some(salt) = self.salts.get(self.salts.len() - 2) {
+            if self.salts.len() > 1 {
+                let salt = &self.salts[self.salts.len() - 2];
                 let now = start_secs + start_instant.elapsed().as_secs() as i32;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/grammers-mtproto/src/mtp/encrypted.rs [hunk 1]
```diff
@@ -1202,10 +1202,12 @@ impl Mtp for Encrypted {
         // Check to see if the next salt can be used already. If it can, drop the current one and,
         // if the next salt is the last one, fetch more.
         if let Some((start_secs, start_instant)) = self.start_salt_time {
-            if let Some(salt) = self.salts.get(self.salts.len() - 2) {
-                let now = start_secs + start_instant.elapsed().as_secs() as i32;
-                if now >= salt.valid_since + SALT_USE_DELAY {
-                    self.salts.pop();
+            if self.salts.len() > 1 {
+                if let Some(salt) = self.salts.get(self.salts.len() - 2) {
```
MaxVerevkin, line 1206: Because of the `if` above, this `.get` must always return `Some`. Let me know if I should leave it as it is, unwrap it, or use `[]` indexing.

Lonami:
CI errors is unrelated; you can rebase from master to solve it.

lib/grammers-mtproto/src/mtp/encrypted.rs
Lonami: Right good catch. The intention with the `get` was precisely to handle the array being too short, but of course the overflow happens sooner. We can definitely use `[]` instead.

PR #246, opened 2024-06-11
remove useless `unsafe impl {Send, Sync}`

MaxVerevkin:
These structs implement `Send + Sync` automatically.


MaxVerevkin:
Now there is no `unsafe` in the repo :)

Lonami:
I've gone ahead and further restricted the use of unsafe in all crates (b9a582a38f678e28e77fe6f2ed072fdfd1e87d3c). Good call removing it.

PR #251, opened 2024-07-21
Some helpful functions for `Media::Photo`, `Media::Document` and `PhotoSize`.

hlf20010508:
- Support `size()` for `Media::Photo` to get the real size of the photo.
- Support `thumbs()` for `Media::Document`.
- Support download to memory for thumb.
- Support get specific type of photo thumb in `VecExt`.


lib/grammers-client/src/types/photo_sizes.rs
```diff
@@ -287,10 +390,19 @@ pub struct PathSize {
 pub trait VecExt {
     /// Helper method to get the largest photo thumb
     fn largest(&self) -> Option<&PhotoSize>;
+
+    /// Helper method to get the specific type of photo thumb
+    ///
+    /// <https://core.telegram.org/api/files#image-thumbnail-types>
+    fn get_type(&self, photo_type: &str) -> Option<&PhotoSize>;
```
Lonami, line 397: I'm not entirely sold on this. How often is this needed, and why can't the user `iter().find()`?

IMO `largest()` is justified because that's the most common scenario.

This API isn't so nice, with a string that must be a single character as input, and documented outside. Feels like it's exposing even more of the raw API, a direction I'd prefer to avoid.

lib/grammers-client/src/types/photo_sizes.rs
```diff
@@ -124,6 +166,45 @@ impl PhotoSize {
...
+            PhotoSize::Path(_) => {
+                // Based on https://core.tlgr.org/api/files#vector-thumbnails
+                let data = self.download_to_memory().await;
+                let mut file = fs::File::create(path).await.unwrap();
+                file.write_all(&data).await.unwrap();
+            }
+        };
+    }
+
+    pub async fn download_to_memory(&self) -> Vec<u8> {
```
Lonami, line 186: This seems very overly specific.

Lonami (changes requested):
I appreciate the PR but I'm not a fan of the seemingly overly-specific functions to download to memory. How often does this happen?

lib/grammers-client/src/types/media.rs
```diff
@@ -155,6 +155,15 @@ impl Photo {
         }
     }
 
+    /// The size of the photo.
+    /// returns 0 if unable to get the size.
+    pub fn size(&self) -> i64 {
```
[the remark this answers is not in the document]
hlf20010508: Because `size()` of `Media::Document` also returns `i64`.

lib/grammers-client/src/types/photo_sizes.rs
hlf20010508: It's ok, I understand your style. I'll revert this.

lib/grammers-client/src/types/photo_sizes.rs
hlf20010508: Downloading to memory is often used when user want to download something and upload it to somewhere immediately, so that useless write and read can be avoid.

Also, to be honest, it's not about whether this is widely used or not, it's about whether your library supports downloading to memory or not. I currently have not found a way to do so without this function, because `DownloadIter::new_from_location` is private. I mean you shouldn't not allow your user to download to memory even if you don't like it right? Still someone will need it.

hlf20010508: And actually I think it's better for the download method in `PhotoSize` to return a `Downloadable`, and allow user to decide whether to use `download_media` or `iter_download`.

lib/grammers-client/src/types/photo_sizes.rs
hlf20010508: By the way, what about change `photo_type` from `String` to a enum, so that user can use that enum easily without going to outside document?

Lonami: Then we need to worry about Telegram potentially removing or adding more stuff.

lib/grammers-client/src/types/photo_sizes.rs
Lonami: Yes, making it a `Downloadable` would be the right way to go, for `iter_download` (and by extension, downloading to memory) to be usable.

lib/grammers-client/src/types/media.rs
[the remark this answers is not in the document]
Lonami: Hm okay, just strange, because we need the cast, but I guess that's how Telegram works.

lib/grammers-client/src/types/photo_sizes.rs
hlf20010508: Sure, I'll have a try.

hlf20010508:
Ok, I've completely migrated download methods in `PhotoSize` to `download_media` and `iter_download`.

The usage is the same as other media
```rust
let downloadable = Downloadable::PhotoSize(photo_size);
let mut download = client.iter_download(&downloadable);
```

I also implemented download for `PhotoSize::Progressive`. It's actually the same as `PhotoSize::Size`. Now users can use `largest()` to get the largest thumb(for photos it's usually progressive) and download it.

And there was a bug in the `download()` of `PathSize`, that the param `path` was shadowed by the variable `mut path`. But it's not needed anymore.

hlf20010508:
The new function `DownloadIter::new_from_photo_size` is ugly, since `request` is not needed, I have to fake a useless one. If you think it's ok to make `request` to `Option<tl::functions::upload::GetFile>`, I'll do that.

Lonami:
From a quick look, as long as it gets the job done, then it's fine.

PR #254, opened 2024-08-10
Optimize Client.next_update signature

dimentyy:
This method does not return Ok(None), so Option is useless


Lonami:
Thanks. I probably made it `Option` to handle the client being disconmected, but indeed that should probably be `Err`, not `None`.

Lonami:
Actually maybe the reason was parity with `Stream`. Not sure.

PR #255, opened 2024-08-11
Add Client.next_raw_update()

dimentyy:
To make it possible, ClientState now stores not the processed updates, but their raw tl forms and associated chat maps.

Instead of converting raw updates to friendly ones immediately after receiving them, Client.next_update now does when needed


Lonami:
Seems reasonable.

PR #256, opened 2024-08-12
Add InlineSend

dimentyy:


lib/grammers-client/examples/dialogs.rs
```diff
@@ -35,7 +34,8 @@ fn prompt(message: &str) -> Result<String> {
     Ok(line)
 }
 
-async fn async_main() -> Result<()> {
+#[tokio::main]
```
Lonami, line 37: Avoiding this macro was deliberate, to prevent `syn+quote` from being compiled in and reduce compile times. At the very least, these changes should be a separate PR, though I'd prefer to keep the old `main` method.

lib/grammers-client/examples/echo.rs
```diff
@@ -68,30 +67,18 @@ async fn async_main() -> Result {
 
     println!("Waiting for messages...");
 
-    // This code uses `select` on Ctrl+C to gracefully stop the client and have a chance to
+    // This code uses `tokio::select!` on Ctrl+C to gracefully stop the client and have a chance to
```
Lonami, line 70: Same here, I know the old code was verbose, but avoiding macros is deliberate.

lib/grammers-client/src/client/updates.rs
```diff
@@ -181,18 +179,9 @@ impl Client {
...
-
-                match select(sleep, step).await {
-                    Either::Left(_) => None,
-                    Either::Right((step, _)) => Some(step),
-                }
-            };
-
-            if let Some(step) = step {
-                step?;
+            tokio::select! {
```
Lonami, line 182: I'd prefer the old code here too, unless it wasn't working properly?

lib/grammers-client/src/types/input_message.rs
```diff
@@ -51,6 +52,14 @@ impl InputMessage {
         self
     }
 
+    /// Whether the media will be inverted.
+    ///
+    /// If inverted, photos, videos, and documents will appear at the bottom and link previews at the top of the message.
```
Lonami, line 57: Oh, that's what it means.

PR #257, opened 2024-08-13
Sending reactions

dimentyy:
Works fine


Lonami:
Appreciate the PRs, but I'd love if there could be a commit per change (one fixing docs, one adding the feature, one renaming). Makes it a lot easier to find why something is the way it is.

Currently I've been squashing to remove stuff like fixing formatting, but separate meaningful commits would be ideal.

PR #263, opened 2024-09-11
feat: impl albums

AndrielFR:
- add `InputMedia` struct.
- add `send_album`, `reply_album` and `respond_album` methods.


lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+    ///
+    /// See also: [`Message::respond_album`], [`Message::reply_album`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # async fn f(chat: grammers_client::types::Chat, client: grammers_client::Client) -> Result<(), Box<dyn std::error::Error>> {
+    /// use grammers_client::InputMedia;
+    ///
+    /// client.send_album(&chat, vec![InputMedia::caption("A album").photo_url("https://www.pixelstalk.net/wp-content/uploads/2016/03/Free-cute-animal-wallpaper-HD.jpg")]).await?;
```
Lonami, line 566: Prefer using a URL like `https://example.com/cat.jpg`.

Any other domain could go down, be hijacked, or a lot of different things. `example.com` is defined for explicit use in examples.

lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+                media.media = Some(
+                    types::Media::from_raw(uploaded)
+                        .unwrap()
+                        .to_raw_input_media()
+                        .unwrap(),
+                );
+            }
+        }
+
+        let first_media = medias.first().unwrap();
```
Lonami, line 605: Oh yeah, I hadn't thought about attributes that only make sense for the entire group. This seems unfortunate. But might be a fair compromise.

lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+                        reply_to_peer_id: None,
+                        quote_text: None,
+                        quote_entities: None,
+                        quote_offset: None,
+                    }
+                    .into()
+                }),
+                schedule_date: None,
+                multi_media: medias
+                    .iter()
```
Lonami, line 626: You could use `into_iter` here…

lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+                    .zip(random_ids.iter())
+                    .map(|(input_media, random_id)| {
+                        let entities = parse_mention_entities(self, input_media.entities.clone());
+                        let raw_media = input_media.media.clone().unwrap();
+
+                        tl::enums::InputSingleMedia::Media(tl::types::InputSingleMedia {
+                            media: raw_media,
+                            random_id: *random_id,
+                            message: input_media.caption.clone(),
```
Lonami, line 635: …which would mean you wouldn't need the `clone` here.

lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+                send_as: None,
+                noforwards: false,
+                update_stickersets_order: false,
+                invert_media: false,
+                quick_reply_shortcut: None,
+                effect: None,
+            })
+            .await?;
+
+        let mut messages = Vec::with_capacity(medias.len());
```
Lonami, line 649: …and if you use `into_iter`, simply take `random_ids.len()` here, which is the same.

lib/grammers-client/src/client/messages.rs
```diff
@@ -546,6 +546,119 @@ impl Client {
...
+        for random_id in random_ids {
+            messages.push(
+                map_random_ids_to_messages(self, &[random_id], updates.clone())
+                    .pop()
```
Lonami, line 653: This function already supports a list of `random_ids`, so I'm not sure why we are calling it one by one.

lib/grammers-client/src/types/input_media.rs
```diff
@@ -0,0 +1,283 @@
...
+        self
+    }
+
+    /// The album identifier to which this album should reply to, if any.
+    ///
+    /// Otherwise, this album will not be a reply to any other.
+    pub fn reply_to(mut self, reply_to: Option<i32>) -> Self {
+        self.reply_to = reply_to;
+        self
+    }
```
Lonami, line 36: Might want to mention that *only* the reply from the *first* media is used.

lib/grammers-client/src/types/input_media.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// For example, this enables you to send a `photo` that can only be viewed for a certain
+    /// amount of seconds before it expires.
+    ///
+    /// Not all media supports this feature.
+    ///
+    /// This method should be called before setting any media, else it won't have any effect.
+    pub fn media_ttl(mut self, seconds: i32) -> Self {
+        self.media_ttl = if seconds < 0 { None } else { Some(seconds) };
+        self
+    }
```
Lonami, line 215: Can this actually be used in albums? What happens if different photos have different TTLs?

lib/grammers-client/src/types/message.rs
```diff
@@ -504,6 +504,17 @@ impl Message {
         self.client.send_message(&self.chat(), message).await
     }
 
+    /// Respond to this message by sending a album in the same chat, but without directly
+    /// replying to it.
+    ///
+    /// Shorthand for `Client::send_album`.
+    pub async fn respond_album(
+        &self,
+        medias: Vec<InputMedia>,
```
Lonami, line 513: I had in mind a slice of `InputMedia`. But perhaps using a `Vec` is okay, as it means we can take ownership and reduce clones. Not sure what's best.

lib/grammers-client/src/client/messages.rs
AndrielFR: oh, im so dumb

lib/grammers-client/src/client/messages.rs
AndrielFR: indeed, it's unfornate

lib/grammers-client/src/types/message.rs
AndrielFR: I think `Vec` is the best option right now

lib/grammers-client/src/types/input_media.rs
AndrielFR: I think the server only uses the first media parameter.

lib/grammers-client/src/types/input_media.rs
AndrielFR: sure

lib/grammers-client/src/client/messages.rs
AndrielFR: sure

lib/grammers-client/src/types/message.rs
Lonami: Not entirely convinced. But we can definitely merge it like this for the time being and see how it feels.

lib/grammers-client/src/client/messages.rs
```diff
@@ -646,17 +646,11 @@ impl Client {
...
-                    .unwrap()
-                    .unwrap(),
-            )
-        }
-
-        Ok(messages)
+        Ok(map_random_ids_to_messages(
+            self,
+            &random_ids,
+            updates.clone(),
```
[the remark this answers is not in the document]
AndrielFR: Oh, I forgot about that one

lib/grammers-client/src/types/message.rs
AndrielFR: Ok.

Lonami:
> Isn't `InputMedia` just a smaller version of `InputMessage`? Why not use `InputMessage`?

Yes, but this was deliberate, because not all messages are valid for album media.

I think a new type is justified, because albums *require* media to be set, while `InputMessage` don't. We would have to add even more validations, and then return `Err`, and the current `Error` type used by the client methods is already complex enough. So I'm not sure it's worth reusing the type.

What I *did* consider however, was using `InputMedia` *for* attaching it to the `InputMessage`. But that's a different problem entirely.

PR #268, opened 2024-10-03
Ability to resolve packed chat by id

MOZGIII:


MOZGIII:
Closing since the updates are not redelivered on restart and the workaround won't work

PR #269, opened 2024-10-03
Add .editorconfig

MOZGIII:
This is recommended to properly autoconfigure the editors for all contributors


Lonami (commented):
Never heard of this. Guess it doesn't hurt, though ideally there would be a pre-commit hook enforcing this.

PR #272, opened 2024-10-07
Catch internal salt responses in mtproto

er-azh:
## Description:
I tried to fix `salt_request_msg_id` to ensure that salt responses are caught internally and don't reach the `Sender`.

## My changes:
- `salt_request_msg_id` is now an `Option<MsgId>` instead of `Option<i64>` to match other types.
- It is assigned the message ID of the new salt request when requesting additional salts.
- A redundant check is removed.
- It gets reset to `None` upon a reset.


Lonami:
Thanks!

Issue #276, opened 2024-10-16
Feature request: serde support


RuofengX:
I am using this awesome crate in my actual work to gather some useful information from open chats and bots.

I want to save the data to database or just simple json-string, like 
- `grammers_client::types::chat::Chat` 
- `grammers_tl_types::types::Message`
- etc.

I found this project seems doesn't support serde, am I right? And if it really not support, why? I noticed that telegram has it's own bytecode ser&de function, but serde is the in-facts standard in rust language. Using serde framework will make this project more flexible.

RuofengX:
Thanks for reply.

I didn't notice such structure is generated. Is that done by `grammers-tl-gen`? If so, could you please tell how the crate works?

RuofengX:
How often does `build.rs` run? Is it just a oneshot task? Or does telegram official often change the tl schema? 

So is it possible to simply add #[derive(Serialize, Deserialized)] in generated files?

Lonami:
`build.rs` runs when its crate or the dependencies of its crate change. So changing the `api.tl`, the generator, or the types package itself, should cause it to re-run.

Lonami:
> `build.rs` runs when its crate or the dependencies of its crate change. So changing the `api.tl`, the generator, or the types package itself, should cause it to re-run.

> telegram official often change the tl schema

Pretty regularly, yes. But grabbing the latest `api.tl` is a manual task.

PR #277, opened 2024-10-18
Add serde support for tl_types.

RuofengX:
Add two lines for each entrance in generated.rs 
- all enums
- all structs
(Did I miss for other things?)

Pass all tests in crate `grammers-tl-types`

### DIFF
--- a/lib/grammers-tl-gen/src/enums.rs
+++ b/lib/grammers-tl-gen/src/enums.rs
@@ -37,7 +37,9 @@ fn write_enum<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
-    writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    if config.impl_serde{
+        writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    }
 
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
     writeln!(
--- a/lib/grammers-tl-gen/src/enums.rs
+++ b/lib/grammers-tl-gen/src/enums.rs
@@ -37,8 +37,11 @@ fn write_enum<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
-    if config.impl_serde{
-        writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    if config.impl_serde {
+        writeln!(
+            file,
+            "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]"
+        )?;
     }
 
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
--- a/lib/grammers-tl-gen/src/enums.rs
+++ b/lib/grammers-tl-gen/src/enums.rs
@@ -37,6 +37,8 @@ fn write_enum<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
+    writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
     writeln!(
         file,
--- a/lib/grammers-tl-gen/src/structs.rs
+++ b/lib/grammers-tl-gen/src/structs.rs
@@ -63,7 +63,9 @@ fn write_struct<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
-    writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    if config.impl_serde{
+        writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    }
 
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
     write!(
--- a/lib/grammers-tl-gen/src/structs.rs
+++ b/lib/grammers-tl-gen/src/structs.rs
@@ -81,11 +81,14 @@ fn write_struct<W: Write>(
 
     writeln!(file)?;
     for param in def.params.iter() {
-        match param.ty {
+        match &param.ty {
             ParameterType::Flags => {
                 // Flags are computed on-the-fly, not stored
             }
-            ParameterType::Normal { .. } => {
+            ParameterType::Normal { ty, .. } => {
+                if config.impl_serde && ty.name.as_str() == "bytes" {
+                    writeln!(file, "{}    #[serde(with = \"serde_bytes\")]", indent)?;
+                }
                 writeln!(
                     file,
                     "{}    pub {}: {},",
--- a/lib/grammers-tl-gen/src/structs.rs
+++ b/lib/grammers-tl-gen/src/structs.rs
@@ -63,6 +63,8 @@ fn write_struct<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
+    writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
     write!(
         file,
--- a/lib/grammers-tl-gen/src/structs.rs
+++ b/lib/grammers-tl-gen/src/structs.rs
@@ -63,8 +63,11 @@ fn write_struct<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
-    if config.impl_serde{
-        writeln!(file, "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]")?;
+    if config.impl_serde {
+        writeln!(
+            file,
+            "{indent}#[derive(serde_derive::Serialize, serde_derive::Deserialize)]"
+        )?;
     }
 
     writeln!(file, "{indent}#[derive(Clone, PartialEq)]")?;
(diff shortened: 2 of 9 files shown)


RuofengX:
Related issue #276

RuofengX:
Whoops, generated code seems encount some lifetime issue.

Lonami:
Can you revert the formatting change to the .toml files out of this PR?

Lonami (changes requested):
I'm not sure I understand why so many crates need the serde dependency. Shouldn't it only be `grammers-tl-types`?

Also, please make it optional. I want to continue offering the crates with serde off by default.

RuofengX:
> I'm not sure I understand why so many crates need the serde dependency. Shouldn't it only be grammers-tl-types?

I tried to add serde code only in `grammers-tl-types` first, all tests pass in `tl-types` crates. 

Then I add my fork of  `grammers-client` in my project, it compile with error saying that the generated `generated.rs`(that long file) missing dependence `serde`. I checked it twice, the error is not from `grammers-tl-types`, it from `grammers-seesion`. Then I added `serde` and `serde-derive` dep in all crates that depend on `grammers-tl-types`. Thus no error again.

I not formillar with build system, but I am looking into it.

---

> Also, please make it optional. I want to continue offering the crates with serde off by default.

Sure, I will add a feature flag in `grammers-client` crate, and add an optional params in `grammers-tl-gen` crate.

RuofengX:
Issue 1 Caused by: I add serde dependence to [dependence] not [build-dependence]

RuofengX:
I think it's done.

User can fetch any raw struct (defined by `grammers_tl_types`) returned by client function after enable the serde feature tag. Structure in `grammers_client` like `Client`,  `Update` still not serializable, because it contains runtime data. Further work need to be done.

RuofengX:
We need more test and docs about this new feature.

lib/grammers-tl-types/src/lib.rs
```diff
@@ -84,19 +86,24 @@ pub use deserialize::{Cursor, Deserializable};
...
 #[derive(Clone, Debug, PartialEq)]
 pub struct RawVec<T>(pub Vec<T>);
 
 /// This struct represents an unparsed blob, which should not be deserialized
 /// as a bytes string. Used by functions returning generic objects which pass
 /// the underlying result without any modification or interpretation.
+#[cfg_attr(feature = "impl-serde", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, PartialEq)]
-pub struct Blob(pub Vec<u8>);
+pub struct Blob(#[cfg_attr(feature = "impl-serde", serde(with = "serde_bytes"))] pub Vec<u8>);
```
RuofengX, line 106: ## Help needed. 

I am not formiliar with tl proto. Is there any other structure that store bytes in .tl definition files? All bytes-like properties need marked by `serde_bytes` for better performence.

lib/grammers-tl-gen/src/enums.rs
```diff
@@ -37,6 +37,10 @@ fn write_enum<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
+    if config.impl_serde{
```
Lonami, line 40: Formatting is going to complain here.

lib/grammers-tl-gen/src/structs.rs
```diff
@@ -63,6 +63,10 @@ fn write_struct<W: Write>(
         writeln!(file, "{indent}#[derive(Debug)]")?;
     }
 
+    if config.impl_serde{
```
Lonami, line 66: And here.

lib/grammers-tl-types/src/lib.rs
Lonami: Yes, `Vec<u8>` are also generated for the TL `bytes` type. If that's the case, we may need to check the field types when generating code.

Might be good to write a `test` that makes sure this works as expected.

Lonami: Yes, `Vec<u8>` are also generated for the TL `bytes` type. If that's the case, we may need to check the field types when generating code.

Might be good to write a `test` that makes sure this works as expected.

lib/grammers-tl-gen/src/enums.rs
RuofengX: true

lib/grammers-tl-types/src/lib.rs
RuofengX: done in [b3573e3](https://github.com/Lonami/grammers/pull/277/commits/b3573e3d178636a25c2986519be1b5712fbebc3f)

lib/grammers-tl-types/Cargo.toml
```diff
@@ -23,6 +23,8 @@ grammers-tl-gen = { path = "../grammers-tl-gen", version = "0.7.0" }
 grammers-tl-parser = { path = "../grammers-tl-parser", version = "1.1.2" }
 
 [dev-dependencies]
+grammers-tl-gen = { path = "../grammers-tl-gen", version = "0.7.0" }
```
RuofengX, line 26: Since new test need to run in compile time, dev-dependencies should add crates used by compile-time.

lib/grammers-tl-types/tests/serde.rs
```diff
@@ -0,0 +1,58 @@
+include!("../build.rs");
+
+#[derive(Debug, Default)]
+struct Scanner {
+    last_line: String,
+}
+impl Scanner {
+    const BYTES_PATTERN: [&str; 2] = ["Vec<u8>", "Option<Vec<u8>>"];
```
RuofengX, line 8: Assert  that there is not any `&'a [u8]` type in `generated.rs`.

lib/grammers-tl-types/tests/serde.rs
Lonami, line 1: This is not a test I had in mind, only [the one from the generator itself](https://github.com/Lonami/grammers/pull/277/files#diff-7c6a6bd6400ff465cb6b568df560f3388659fe24a230b8551b4c9c880553e432R125).

I'm sure this was useful when making the PR, but I'm not sure we need it. I doubt Telegram will add any byte storage other than `bytes` and `strings`.

So I'm inclined to say the other test is enough. And if we delete this, we wouldn't need the build-dependencies, right? Or is there a strong reason to keep this that I'm not seeing?

Lonami:
First PRs require me to click the Approve button every time to have CI run, so it might take a while until I realize I had to click it again.

RuofengX:
> First PRs require me to click the Approve button every time to have CI run, so it might take a while until I realize I had to click it again.

Seems like unit test file is not latest. I run cargo test in local successful. Checking.

lib/grammers-tl-types/tests/serde.rs
RuofengX: I should say yes, this check is a duplicate with the test in grammers-tl-gen. 
I will remove it, and roll back dependence.

RuofengX: > doubt Telegram will add any byte storage other than bytes and strings.

I think this is not a problem. 

1. serde_bytes only support Vec<u8>, &[u8], [u8;N], their borrow ed form and their Option<T> form.

2. The serde_bytes mark will added to attributes with tl's bytes type. This mark works if tl.bytes finally parsed as a rust type that serde_bytes support. Currently it's Vec<u8>, so the mark works. 

4. On the other, if tl schema file changed and has other types that contains binary data, but not using bytes, serde_bytes mark will not added to it's rust form, thus serde will treat them as a normal vector or other thongs, but serde could still run, with little reduced performance.

5. We need great effort s to fix issue in point 4. 
This is the ideal logic: check the generated code lines, if one line contains Vec<u8> or other things that serde_bytes support just add serde_bytes mark before this line.   
But consider 2 scenarios. S1: Simple string match will accidentally add marks to Vec<Vec<u8>>, which serde_bytes doesn't support.   
S2. Things like i128 in tl is interpreted to  [u8; 16] in rust, which serde_bytes does support, this situation should be excluded. 

S1 and S2 makes such string match method really difficult. Current type detect method is a trade off of (very little) performance loss with programming complexities.

RuofengX: We have some Vec<Vec<u8>> type now in `generated.rs`, and it is not bytes, like `CodeSetting.logout_tokens`, it is a Vec<Vec<u8>>.  
First vector is a list, while secode is a bytes.  
Maybe we could add a type alias in generate logic, or use 3-rd party dependence like `bytes` from tokio team to support serde and nested type more easily.

Anyway, I fix the broken test, it is because the blank space in generated code. And revert to the commit that no new duplicate test in `grammers-tl-types`. 

What else before merge?

RuofengX:
We have some `Vec<Vec<u8>>` type now in `generated.rs`, and it cannot get parsed by `serde_bytes`, like `CodeSetting.logout_tokens`, it is a `Vec<Vec<u8>>`.  
First vector is a list, while second is a bytes.  

I don't know whether `Vec<#[serde(with = "serde_bytes")] Vec<u8>>` will work. Moreover current inject point cannot capture the exact begin of a bytes type, it can only add mark before definition line.  The nested write processing wipes the boundry between list and bytes.  

Plan 1. If `Vec<#[serde(with = "serde_bytes")] Vec<u8>>` works, we could write serde_bytes mark just before `bytes` type parsed. I can foresee a lot refactor to pass params.

Plan 2. Maybe we could add a type alia in generated file to distinguish bytes and vector in generate logic.

Plan 3. We could use 3-rd party dependence like `bytes` from tokio team to support serde and nested type more easily. But this will bring more dependences and limit downstream user's choice.

Plan 4. Leave this as a flaw, do nothing about it.

Anyway, I fix the broken test, it is because the blank space in generated code. And revert to the commit that no new duplicate test in `grammers-tl-types`.

Lonami:
> Plan 4. Leave this as a flaw, do nothing about it.

I think this is fine. Needing to serialize types like that seems extremely unlikely. As long as it works, even if it's inefficient, I think it's fine.

Lonami (approved):

Issue #191, opened 2023-07-12
Answer a reply_markup from a BOT message


felipepimsil:
Is there a way to answer a reply_markup? I connected with my user and wanted to answer some buttons from a BOT message, but I didn't find some method.

Lonami:
Hi, this is not an issue with the library but rather a question. The issues section in GitHub is only to report bugs or enhancements in the library itself. If you have usage questions, you should ask those in [StackOverflow](https://stackoverflow.com/questions/ask?tags=telethon) or [the @TelethonChat group](https://t.me/TelethonChat).

RuofengX:
Client currently don't have a reply markup method.

Lonami:
When I posted my comment, I thought I was in the Telethon repository.

Either way, since the issue was posted, it's now possible:
https://github.com/Lonami/grammers/blob/8997a2dadc64b9ee7ee4d03b2f6eaba15d44f00c/lib/grammers-client/src/types/input_message.rs#L81

RuofengX:
> When I posted my comment, I thought I was in the Telethon repository.
> 
> Either way, since the issue was posted, it's now possible:
> https://github.com/Lonami/grammers/blob/8997a2dadc64b9ee7ee4d03b2f6eaba15d44f00c/lib/grammers-client/src/types/input_message.rs#L81

I think there is a mistake.  

reply_markup function send a button in message.   
We still can't press the button that peer send. 


For example, I am talking with a bot, received a message with button on it, and want my client to press a button behind the msg sent by bot.

Lonami:
You can `client.invoke` that request, but you're right a friendly API is missing.

Ideally the `Message` custom type would have a `.buttons()` that can be `.click()`ed.

PR #279, opened 2024-10-21
support callback button

RuofengX:
#191

### DIFF
--- a/lib/grammers-client/src/types/update/callback_query.rs
+++ b/lib/grammers-client/src/types/update/callback_query.rs
@@ -103,7 +103,7 @@ impl CallbackQuery {
     }
 
     /// Answer the callback query.
-    pub fn answer(&self) -> Answer {
+    pub fn answer(&self) -> Answer<'_> {
         let query_id = match &self.raw {
             tl::enums::Update::BotCallbackQuery(update) => update.query_id,
             tl::enums::Update::InlineBotCallbackQuery(update) => update.query_id,
--- a/lib/grammers-client/src/types/update/update.rs
+++ b/lib/grammers-client/src/types/update/update.rs
@@ -10,7 +10,7 @@ use std::sync::Arc;
 
 use super::{CallbackQuery, InlineQuery, InlineSend, Message, MessageDeletion, Raw};
 use crate::types::Message as Msg;
-use crate::{ChatMap, Client};
+use crate::{ChatMap, Client, utils};
 use grammers_session::State;
 use grammers_tl_types as tl;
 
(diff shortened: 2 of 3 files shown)


lib/grammers-client/src/types/message.rs
```diff
@@ -388,6 +388,50 @@ impl Message {
         self.raw.reply_markup.clone()
     }
 
+    /// Collect all inline callback buttons. Those buttons often appeare in bot chat.
+    /// Buttons could be clicked by [`click_callback_button`]
+    pub fn callback_buttons(&self) -> Option<Vec<tl::types::KeyboardButtonCallback>> {
```
RuofengX, line 393: Only handle KeyboardButtonCallback, other KeyboardButton types cannot handle.

lib/grammers-client/src/types/message.rs
```diff
@@ -388,6 +388,50 @@ impl Message {
...
+
+        Some(rtn)
+    }
+
+    pub async fn click_callback_button(
+        &self,
+        button: &tl::types::KeyboardButtonCallback,
+    ) -> Result<()> {
+        if button.requires_password {
+            unimplemented!()
```
RuofengX, line 419: Don't know how to deal with password.

lib/grammers-client/src/types/message.rs
```diff
@@ -388,6 +388,50 @@ impl Message {
...
+        } else {
+            let rtn = self
+                .client
+                .invoke(&tl::functions::GetBotCallbackAnswer {
+                    game: false,
```
RuofengX, line 424: Don't know how to deal with game flag.

lib/grammers-client/src/types/message.rs
RuofengX, line 1: Still lot works to do. Described in file comment.

RuofengX (commented):
Here list a todo-list.

lib/grammers-client/src/types/message.rs
Lonami: I was thinking more to return a custom type of buttons, so that `.click()` could be called directly on the one you need.

By the way, might be good to call this "inline buttons". The term in your code is used by the API, but I believe the other term is more common in the applications? Not sure what the bot API calls it.

lib/grammers-client/src/types/message.rs
Lonami: Yeah it's a bit of a mess. I'd take the implementation from Telethon.

lib/grammers-client/src/types/message.rs
Lonami: Likewise.

lib/grammers-client/src/types/message.rs
```diff
@@ -388,6 +388,50 @@ impl Message {
...
+        let mut rtn = Vec::new();
+        if let tl::enums::ReplyMarkup::ReplyInlineMarkup(markup) = reply_markup {
+            for row in markup.rows.iter() {
+                let tl::enums::KeyboardButtonRow::Row(row) = row;
+                for b in row.buttons.iter() {
+                    match b {
+                        tl::enums::KeyboardButton::Callback(callback_b) => {
+                            rtn.push(callback_b.clone());
+                        }
+                        _ => (),
```
Lonami, line 405: This means the position in the returned array will not match the one users see in the application. Which would be really confusing.

xximwon (approved):

PR #283, opened 2024-10-25
Add a method to allow using TL types for media in InputMessage 

er-azh:
Adds two new methods to `InputMessage` , `photo_id` and `document_id`. These new methods allow the user to include media in their messages using `InputPhoto` and `InputDocument` without having to copy the media from a message or re-uploading.

### DIFF
--- a/lib/grammers-client/src/types/input_message.rs
+++ b/lib/grammers-client/src/types/input_message.rs
@@ -187,6 +187,20 @@ impl InputMessage {
         self
     }
 
+    /// Include a media in the message using the raw TL types.
+    ///
+    /// You can use this to send any media using the raw TL types that don't have
+    /// a specific method in this builder such as Dice, Polls, etc.
+    ///
+    /// This can also be used to send media with a file reference, see `InputMediaDocument`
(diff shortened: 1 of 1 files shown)


Lonami (commented):
I'm not convinced on the naming. Would be nice to accept a generic for either `Uploaded` or the input version.

er-azh:
Hmmm, what about adding something like this and then adding `into_document` and `into_photo` to `Uploaded` instead?
```rust
pub fn media<T: Into<tl::enums::InputMedia>>(mut self, media: T) -> Self {
    self.media = Some(media.into());
    self
}
```
The issue with using a generic parameter for the current `document` and `photo` methods is that I can't think of a way to differentiate between the media types (eg. `InputMediaPhoto` and `InputMediaDocument`) while accepting both uploaded and forwarded variants of `InputMedia` (eg. `InputMediaPhoto` and `InputMediaUploadedPhoto`) without adding at least one unnecessary wrapper type.

Lonami:
I was thinking of two traits, `IntoInputPhoto` and `IntoInputDocument`, both implemented for `Uploaded`, but only one for the raw types.

er-azh:
I've updated everything based on the telegram discussion.

PR #285, opened 2024-11-04
Add support to inline callback query

AndrielFR:


lib/grammers-client/src/types/callback_query.rs
```diff
@@ -142,11 +172,19 @@ impl<'a> Answer<'a> {
...
-        self.query
-            .client
-            .edit_message(chat, msg_id, new_message)
-            .await
+        if let Some(ref msg_id) = self.query.inline_msg_id {
+            self.query
+                .client
+                .edit_inline_message(msg_id.clone(), new_message)
+                .await
+                .map(|_| ())
```
Lonami, line 180: IIRC `map(drop)` was preferred as this looks like a toilet.

lib/grammers-client/src/types/callback_query.rs
```diff
@@ -44,6 +45,30 @@ impl CallbackQuery {
...
+        }
+    }
+
+    pub fn from_inline_raw(
+        client: &Client,
+        query: tl::types::UpdateInlineBotCallbackQuery,
+        chats: &Arc<types::ChatMap>,
+    ) -> Self {
+        Self {
+            raw: tl::types::UpdateBotCallbackQuery {
```
Lonami, line 58: Not sure how I feel about "lying" about the original raw type (but I can't remember if we do it elsewhere, and updates are quite similar, so this is OK).

lib/grammers-client/src/client/bots.rs
```diff
@@ -153,27 +153,29 @@ impl Client {
...
     pub async fn edit_inline_message<M: Into<InputMessage>>(
         &self,
         message_id: tl::enums::InputBotInlineMessageId,
         input_message: M,
     ) -> Result<bool, InvocationError> {
         let message: InputMessage = input_message.into();
         let entities = parse_mention_entities(self, message.entities);
-        let dc_id = message_id.dc_id();
         let result = self
-            .invoke_in_dc(
```
Lonami, line 165: Would be good to make sure with messages with media in other DCs.

lib/grammers-client/src/types/callback_query.rs
AndrielFR: kkkk, sure

lib/grammers-client/src/client/bots.rs
AndrielFR: Done

lib/grammers-client/src/types/callback_query.rs
AndrielFR: ok

Lonami (approved):

PR #287, opened 2024-11-06
deps: trust-dns is no longer maintained, move into hickory-dns

yhx-12243:
See main page of https://github.com/bluejekyll/trust-dns.


Lonami:
I'd like to remove direct support for proxies eventually (and instead provide a more generalised solution).

But this should be good for now.

Issue #274, opened 2024-10-08
Chat name and reconnection issues


sigeshuo:
```rust
use chrono::Local;
use fern::colors::{Color, ColoredLevelConfig};
use futures_util::future::{select, Either};
use grammers_client::session::Session;
use grammers_client::types::{chat, message};
use grammers_client::{types, Client, Config, InitParams, SignInError, Update};
use grammers_mtsender::{FixedReconnect, ReconnectionPolicy};
use grammers_session::PackedType;
use grammers_session::PackedType::Chat;
use log::{error, info, warn, LevelFilter};
use simple_logger::*;
use std::io::Write;
use std::io::{self, BufRead as _, Write as _};
use std::ops::ControlFlow;
use std::pin::pin;
use std::time::Duration;
use std::time::SystemTime;
use tokio::{runtime, task};

type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;

const SESSION_FILE: &str = "tdata";

struct AutoReconnectPolicy;

impl ReconnectionPolicy for AutoReconnectPolicy {
    fn should_retry(&self, attempts: usize) -> ControlFlow<(), Duration> {
        let duration = u64::pow(2, attempts as _);
        warn!("Auto reconnect {} time, next reconnect time {} seconds later",attempts, duration);
        ControlFlow::Continue(Duration::from_secs(duration))
    }
}

fn prompt(message: &str) -> Result<String> {
    let stdout = io::stdout();
    let mut stdout = stdout.lock();
    stdout.write_all(message.as_bytes())?;
    stdout.flush()?;

    let stdin = io::stdin();
    let mut stdin = stdin.lock();

    let mut line = String::new();
    stdin.read_line(&mut line)?;
    Ok(line)
}

fn init_logger() {
    // 配置不同日志级别的颜色
    let colors = ColoredLevelConfig::new()
        .info(Color::Green)
        .warn(Color::Yellow)
        .error(Color::Red)
        .debug(Color::Blue);

    fern::Dispatch::new()
        // 设置最低日志级别
        .level(LevelFilter::Info)
        // 给不同的日志级别上色
        .format(move |out, message, record| {
            let timestamp = Local::now().format("%Y-%m-%d %H:%M:%S");
            out.finish(format_args!(
                "[{}] [{}] {} - {}",
                timestamp,
                colors.color(record.level()),
                "Telegram Client",
                message
            ))
        })
        // 日志输出到标准输出
        .chain(io::stdout())
        .apply()
[...]

Issue #212, opened 2023-12-07
message.chat().username() always None


AIPACMan:
Is this the correct behavior when a user writes for the first time, then `message.chat().username()` == Some('Username'). But when the same user writes subsequent messages, message.chat().username() is always None

Lonami:
Sounds about right. Telegram doesn't include all sender information because it assumes you already have it. Only way to solve it would be to either add a cache to grammers (not happening, that has a lot of other problems to solve) or always make a request (wasteful, some users won't need it).

Therefore, there's nothing to do here. Code should rely on the ID, not the username.

PR #292, opened 2024-11-22
Remove sentinel values in names

rsxrwscjpzdzwpxaujrr:
Functions `Chat::name`, `Chat::User::first_name`, and `Chat::Group::title` have sentinel values, occasionally returning empty strings, which is a bad code practice. The docs also don't mention all the cases when the strings are going to be empty.

Telegram doesn't include all sender information because it assumes you already have it.
_Originally posted by @Lonami in https://github.com/Lonami/grammers/issues/212#issuecomment-1844791353_

This creates confusion, leading to faulty code, as in #274 for example. Using `Option` will make it so users can't accidentally forget to check if the returned string is empty.

### DIFF
--- a/lib/grammers-client/examples/dialogs.rs
+++ b/lib/grammers-client/examples/dialogs.rs
@@ -93,7 +93,7 @@ async fn async_main() -> Result<()> {
     println!("Showing up to {} dialogs:", dialogs.total().await?);
     while let Some(dialog) = dialogs.next().await? {
         let chat = dialog.chat();
-        println!("- {: >10} {}", chat.id(), chat.name());
+        println!("- {: >10} {}", chat.id(), chat.name().unwrap_or_default());
     }
 
     if sign_out {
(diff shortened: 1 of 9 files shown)


Lonami:
My thinking here was that `String` already has an empty state, and the `None` vs `String::default()` distinction isn't meaningful, because Telegram does not return or allow setting empty strings as names.

But it is true that the `Option` makes it far more explicit, although also more annoying in the common (?) case.

rsxrwscjpzdzwpxaujrr:
It's more annoying to use `Option` when you don't check if the returned string is empty, which is the case for examples and tests, as **none of them ever check**, which is a buggy code (and btw that's a good example why sentinel values are bad). echo.rs prints "Responding to " without a name most of the time. Telegram not sending names when you already have them basically means the common case is when it doesn't send the name, so real-world applications must have some kind of cache of the names and check if the string is empty every time. And when you do check, it is becoming more annoying to have `&str` with sentinel values than `Option<&str>`, as you don't have all the sugar that `Option` provides and need to do something like `chat.name().is_empty().not().then(|| chat.name())` to be able to, for example, match on it.

The meaning of `None` vs `String::default()` is to make it more convenient to write good code and less convenient to write bad code. Using empty strings as sentinel values resembles the usage of null values ("hey, a memory address can't be 0, let's use it when we don't have the value") and is not rust-way.

The ultimate one to blame for the annoyance here is Telegram not sending the names with every message.

PR #295, opened 2024-12-20
Migrating to `futures::Stream` for all `iter_*` method

YouKnow-sys:
In this PR I plan to rewrite all `iter_*` methods to use [`futures::Stream`](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) instead of manual iterator implementation.
Using [`futures::Stream`](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) could give us some very good advantages, for example there are tons of useful traits in `futures` that we can then use with our `Stream`'s and its also overall a more standard way then what we are doing right now.

# Todo:

* [x] `iter_download` => `stream_download`
* [x] `iter_dialogs` => `stream_dialogs`
* [x] `iter_participants` => `stream_participants`
* [x] `iter_profile_photos` => `stream_profile_photos`
* [x] `iter_messages` => `stream_messages`
* [x] `search_messages` => `search_messages`
* [ ] `search_all_messages` => `search_all_messages`
* [ ] Improve the overall code

# Questions:
1. should we really change the name of methods to `stream_*`? or is there better names
2. should we re-export `futures` crate?
3. using `pin_mut` macro everywhere doesn't seems like a good idea, what else should we do? should we implement custom future for lower level types?

### DIFF
--- a/lib/grammers-client/Cargo.toml
+++ b/lib/grammers-client/Cargo.toml
@@ -22,6 +22,7 @@ serde = ["grammers-tl-types/impl-serde"]
 
 [dependencies]
 chrono = "0.4.38"
+futures = "0.3.31"
 futures-util = { version = "0.3.30", default-features = false, features = [
     "alloc"
 ] }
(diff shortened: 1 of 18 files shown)


Lonami:
Closing since all commits are now in `master`. Thanks.

PR #298, opened 2024-12-20
[WASM-1] Implement the obfuscated MTProto transport

er-azh:
This PR implements support for [transport obfuscation](https://core.telegram.org/mtproto/mtproto-transports#transport-obfuscation). It is the first step towards a web port.

## Changes I had to make to make this feasible:
- `Transport::unpack()` now takes in a mutable reference to the buffer to do the decryption in-place.
- ~`Transport::unpack()` now returns `Result<Vec<UnpackedOffset>>`. This is needed since a read might, and frequently will, contain multiple transport payloads.~
- ~`Transport::unpack()` now returns an empty Vec instead of `Error::MissingBytes` when there's more data needed.~
- `ctr` is a new dependency in `grammers-mtsender`.
- `Transport::unpack()` must now be called with the same buffer and the buffer must be kept between calls.

~While changing the return type of unpack to a `Vec` doesn't also seem good, it was the only compromise I could think of to allow this transport to work without significant performance impact while also keeping the decryption contained inside of the transport implementation.~

I'm now storing the decryption state inside the transport itself. So should have almost zero overhead other than the encryption itself

### DIFF
--- a/lib/grammers-crypto/src/obfuscated.rs
+++ b/lib/grammers-crypto/src/obfuscated.rs
@@ -1,0 +1,43 @@
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use aes::cipher::{generic_array::GenericArray, KeyIvInit, StreamCipher};
+
+/// This implements the AES-256-CTR cipher used by Telegram to encrypt data
+/// when using the obfuscated transport.
+///
+/// You're not supposed to use this directly, You're probably looking for the
+/// actual implementation in `grammers-mtproto`.
+pub struct ObfuscatedCipher {
+    rx: ctr::Ctr128BE<aes::Aes256>,
+    tx: ctr::Ctr128BE<aes::Aes256>,
+}
+
+impl ObfuscatedCipher {
+    pub fn new(init: &[u8; 64]) -> Self {
+        let init_rev = init.iter().copied().rev().collect::<Vec<_>>();
+        Self {
+            rx: ctr::Ctr128BE::<aes::Aes256>::new(
+                GenericArray::from_slice(&init_rev[8..40]),
+                GenericArray::from_slice(&init_rev[40..56]),
+            ),
+            tx: ctr::Ctr128BE::<aes::Aes256>::new(
+                GenericArray::from_slice(&init[8..40]),
+                GenericArray::from_slice(&init[40..56]),
+            ),
+        }
+    }
+
+    pub fn encrypt(&mut self, buffer: &mut [u8]) {
+        self.tx.apply_keystream(buffer);
+    }
+
+    pub fn decrypt(&mut self, buffer: &mut [u8]) {
+        self.rx.apply_keystream(buffer);
+    }
+}
--- a/lib/grammers-mtproto/src/transport/abridged.rs
+++ b/lib/grammers-mtproto/src/transport/abridged.rs
@@ -5,7 +5,7 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-use super::{Error, Transport, UnpackedOffset};
+use super::{Error, Tagged, Transport, UnpackedOffset};
 use grammers_crypto::DequeBuffer;
 
 /// The lightest MTProto transport protocol available. This is an
@@ -109,6 +109,13 @@ impl Transport for Abridged {
     fn reset(&mut self) {
         log::info!("resetting sending of header in abridged transport");
         self.init = false;
+    }
+}
+
+impl Tagged for Abridged {
+    fn init_tag(&mut self) -> [u8; 4] {
+        self.init = true;
+        [0xef, 0xef, 0xef, 0xef]
     }
 }
 
--- a/lib/grammers-mtproto/src/transport/abridged.rs
+++ b/lib/grammers-mtproto/src/transport/abridged.rs
@@ -63,7 +63,7 @@ impl Transport for Abridged {
         }
     }
 
-    fn unpack(&mut self, buffer: &[u8]) -> Result<UnpackedOffset, Error> {
+    fn unpack(&mut self, buffer: &mut [u8]) -> Result<UnpackedOffset, Error> {
         if buffer.is_empty() {
             return Err(Error::MissingBytes);
         }
@@ -160,7 +160,7 @@ mod tests {
         let mut transport = Abridged::new();
         let mut buffer = DequeBuffer::with_capacity(1, 0);
         buffer.extend([1]);
-        assert_eq!(transport.unpack(&buffer[..]), Err(Error::MissingBytes));
+        assert_eq!(transport.unpack(&mut buffer[..]), Err(Error::MissingBytes));
     }
 
     #[test]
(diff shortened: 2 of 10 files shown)


lib/grammers-crypto/src/obfuscated.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_revert() {
+        let mut cipher = ObfuscatedCipher::new(&[0x69; 64]);
+        let orignial = [0x42; 128];
+        let mut buffer = orignial.clone();
```
Lonami, line 66:
```suggestion
        let original = [0x42; 128];
        let mut buffer = original.clone();
```

lib/grammers-crypto/src/obfuscated.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        cipher.decrypt(&mut buffer);
+        cipher.undo_decrypt(&mut buffer);
+        assert_eq!(&buffer, &orignial);
```
Lonami, line 70:
```suggestion
        assert_eq!(&buffer, &original);
```

lib/grammers-mtproto/src/transport/abridged.rs
```diff
@@ -99,11 +99,11 @@ impl Transport for Abridged {
         let header_len = header_len as usize;
         let len = len as usize;
 
-        Ok(UnpackedOffset {
+        Ok(vec![UnpackedOffset {
```
Lonami, line 102: I'm not sure forcing an allocation like this, as cheap as it might be in practice, is worth it.

lib/grammers-mtproto/src/transport/abridged.rs
```diff
@@ -112,6 +112,13 @@ impl Transport for Abridged {
     }
 }
 
+impl Tagged for Abridged {
+    fn obfuscated_tag(&mut self) -> [u8; 4] {
+        self.init = true;
```
Lonami, line 117: A getter mutating itself doesn't feel right.

Maybe the method should be called `init_tag` or similar, if we really need that. Or otherwise just split the tag step from the init step.

lib/grammers-mtproto/src/transport/obfuscated.rs
```diff
@@ -0,0 +1,131 @@
...
+    inner: T,
+    head: Option<[u8; 64]>,
+    cipher: ObfuscatedCipher,
+}
+
+const FORBIDDEN_FIRST_INTS: [[u8; 4]; 7] = [
+    [0x48, 0x45, 0x41, 0x44], // "HEAD"
+    [0x50, 0x4f, 0x53, 0x54], // "POST"
+    [0x47, 0x45, 0x54, 0x20], // "GET "
+    [0x4f, 0x50, 0x54, 0x49], // "OPTI"
```
Lonami, line 36: Does this read better? Not sure. Either is probably fine.

```suggestion
    [b'H', b'E', b'A', b'D'], // HTTP HEAD
    [b'P', b'O', b'S', b'T'], // HTTP POST
    [b'G', b'E', b'T', b' '], // HTTP GET
    [b'O', b'P', b'T', b'I'], // HTTP OPTIONS
```

Lonami (commented):
IIRC at some point `Error::MissingBytes` also reported how many bytes we were expecting. I probably removed it because it wasn't needed, but now it is.

If we bring that back, then the obfuscated transport could decrypt the first 4 bytes (which contain the entire header for all supported transports), attempt to unpack, and from the reported error we would know how much more we need to decrypt.

Then we can keep a flag "have we decrypted the header yet" (or better yet, store an `Option` with the last-requested size) to avoid doing the decryption on that again. And once we have enough bytes, we decrypt the rest and return that.

Now we can go back to returning a single `Result<Offset>` instead of a `Vec` with the allocations that implies. And it should also mean we don't need to undo decryptions.

WDYT?

lib/grammers-mtproto/src/transport/abridged.rs
er-azh: I've renamed it to `init_tag`.

lib/grammers-mtproto/src/transport/obfuscated.rs
er-azh: Yeah, this is better.

er-azh:
@Lonami 
I've amended my commit and implemented what we discussed in telegram (storing the decryption state inside of the transport itself).
I've also applied all the suggestions you mentioned in the review.

lib/grammers-mtproto/src/transport/mod.rs
```diff
@@ -91,8 +93,17 @@ pub trait Transport {
     fn pack(&mut self, buffer: &mut DequeBuffer<u8>);
 
     /// Unpacks the input buffer in-place.
-    fn unpack(&mut self, buffer: &[u8]) -> Result<UnpackedOffset, Error>;
+    /// Subsequent calls to `unpack` should be made with the same buffer only
+    /// with the old data removed and new data appended.
```
Lonami, line 97: > with the old data removed and new data appended

This part seems misleading. The old data will still be there until `unpack` returns an offset. Once the offset is returned, the processed data is removed.

Maybe this reads better, or maybe you can think of better wording:

```suggestion
    /// Subsequent calls to `unpack` should be made with the same buffer,
    /// with the data on the ranges from previous `UnpackedOffset` removed.
```

er-azh: Done.

lib/grammers-mtproto/src/transport/mod.rs
```diff
@@ -91,8 +93,17 @@ pub trait Transport {
     fn pack(&mut self, buffer: &mut DequeBuffer<u8>);
 
     /// Unpacks the input buffer in-place.
-    fn unpack(&mut self, buffer: &[u8]) -> Result<UnpackedOffset, Error>;
+    /// Subsequent calls to `unpack` should be made with the same buffer,
+    /// with the data on the ranges from previous `UnpackedOffset` removed.
+    fn unpack(&mut self, buffer: &mut [u8]) -> Result<UnpackedOffset, Error>;
```
MOZGIII, line 98: There should probably be an explainer on why the `buffer` is `mut` here... As I was wondering.

lib/grammers-mtproto/src/transport/obfuscated.rs
```diff
@@ -0,0 +1,111 @@
...
+        self.inner.pack(buffer);
+        self.cipher.encrypt(buffer.as_mut());
+        if let Some(head) = self.head.take() {
+            buffer.extend_front(&head);
+        }
+    }
+
+    fn unpack(&mut self, buffer: &mut [u8]) -> Result<UnpackedOffset, Error> {
+        if buffer.len() < self.decrypt_tail {
+            panic!("buffer is smaller than what was decrypted");
```
MOZGIII, line 88: Should this really be a panic? If so - there should be an explainer in the comments as to where this is checked and why this case can't be handled (i.e. we should crash the whole app here).

lib/grammers-mtproto/src/transport/mod.rs
Lonami: `pack` takes a mutable input so in my view it's reasonable for `unpack` to do so too.

lib/grammers-mtproto/src/transport/obfuscated.rs
er-azh: Yes, this should really panic. Reaching here means the user has messed up their buffer in an unexpected way. And there'd be a panic further down anyways, so it's better to provide a user friendly message than to crash with an indexing error.

There's also an explainer in the comments of `Transport::unpack()` and this will never happen when used with `grammers-mtsender`. It's only `panic!(...)` and not `unreachable!(...)` because the crate can be used standalone.

lib/grammers-mtproto/src/transport/mod.rs
MOZGIII: By why would we alter the buffer itself? Isn't it supposed to be the read buffer?

Lonami: Both packing unpacking occur in-place. None of the previous transports had to modify the buffer during unpack, but obfuscated does. The caller doesn't read from the buffer contents until the calls succeed.

Lonami: (Forgot to add, but that said, I'm happy to have the documentation improved to leave this in a better place than a random GitHub discussion.)

PR #301, opened 2024-12-23
[WASM-4] Clean up mtsender, ws support and small changes to make it usable

er-azh:
After this PR users should be able to use grammers in the browser! We still have a lot of testing and documenting to do but that's for another PR.

## Changes:
- Moved `NetStream` into a new mod called `net` and separate net code from the actual mtsender implementation.
- Implement WebSocket support using `ws_stream_wasm`.
- Declare a new type for the transport type in client to allow changing it using cfg.
- Use padded intermediate with obfuscation as the transport when targeting web.
- Enable the `js` feature of `getrandom` to avoid getting compile errors.
- Add a compile error when the user tries to build with the `fs` or `proxy` features but is targeting web.

### DIFF
--- a/lib/grammers-client/src/client/net.rs
+++ b/lib/grammers-client/src/client/net.rs
@@ -39,6 +39,21 @@ const DC_ADDRESSES: [(Ipv4Addr, u16); 6] = [
     (Ipv4Addr::new(91, 108, 56, 190), 443),
 ];
 
+/// WebSocket addresses to Telegram datacenters, where the index into this array
+/// represents the data center ID.
+///
+/// The addresses were obtained from the official docs.
+/// See [URI Format](https://core.telegram.org/mtproto/transports#uri-format).
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+const WS_ADDRESSES: [&str; 6] = [
+    "",
+    "wss://pluto.web.telegram.org/apiws",
+    "wss://venus.web.telegram.org/apiws",
+    "wss://aurora.web.telegram.org/apiws",
+    "wss://vesta.web.telegram.org/apiws",
+    "wss://flora.web.telegram.org/apiws",
+];
+
 const DEFAULT_DC: i32 = 2;
 
 pub(crate) async fn connect_sender(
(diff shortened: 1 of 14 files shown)


MOZGIII:
This is great! I have previously experimented with making mtsender work with traits for transport, for it to be possible to provide your own custom networking, and to make it pluggable - but it was more of an experimental branch than a real implementation. This change here would significantly simplify the work needed to abstract the transport away behind a trait.

Great stuff!

lib/grammers-client/src/client/net.rs
```diff
@@ -38,18 +39,67 @@ const DC_ADDRESSES: [(Ipv4Addr, u16); 6] = [
...
+                ServerAddr::Tcp { address: tcp_addr }
+            };
+
+            addr
+        };
+
+        #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+        let addr = ServerAddr::Ws {
+            address: WS_ADDRESSES[dc_id as usize].to_string(),
+        };
```
MOZGIII, line 100: Question: can't a WS connection be proxied too?

So far, it looks like we should have something like `TcpTransport`. `WebSocketTransport` and `Proxied<T: Transport>` and `enum AnyTransport` with all the supported variants to cover all the permutations properly without creating too much mess.

er-azh: > can't a WS connection be proxied too?
No, we're in a browser. We don't have access to connection properties.

lib/grammers-client/src/client/net.rs
```diff
@@ -52,6 +67,7 @@ pub(crate) async fn connect_sender(
     let addr: ServerAddr = if let Some(ref sa) = config.params.server_addr {
         sa.clone()
     } else {
+        #[cfg(all(not(all(target_arch = "wasm32", target_os = "unknown")),))]
```
Lonami, line 79: Outer `all` seems redundant, unless we're missing a `feature = "proxy"` as we do in sender's `net/mod`.

er-azh: You're right. I missed that when moving both code blocks into a single one. I'll fix it with a force push.

lib/grammers-client/src/client/net.rs
MOZGIII: Is the only use-case for ws is the browser? I can see it used outside of the browser too, in which case it would allow proxying at our end

er-azh: The purpose of this PR is only browser support as I don't have any use for ws in platforms other than in a browser.

PR #304, opened 2025-01-10
allow set max_id and min_id for MessageIter and SearchIter

amone-bit:
allow set max_id and min_id for MessageIter and SearchIter to filter messages like telethon.

### DIFF
--- a/lib/grammers-client/src/client/messages.rs
+++ b/lib/grammers-client/src/client/messages.rs
@@ -220,6 +220,16 @@ impl MessageIter {
         self
     }
 
+    pub fn min_id(mut self, min_id: i32) -> Self {
+        self.request.min_id = min_id;
+        self
+    }
+
+    pub fn max_id(mut self, max_id: i32) -> Self {
+        self.request.max_id = max_id;
+        self
+    }
+
     /// Determines how many messages there are in total.
     ///
     /// This only performs a network call if `next` has not been called before.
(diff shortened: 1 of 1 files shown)


Lonami:
Why is the existing `offset_id` and breaking out of the loop early not enough?

Lonami:
I've discussed this before and also argued it wasn't needed. But I may reconsider if more people need it.

One of my worries is that some offsets seem to not always work, and exposing them here would mean people would expect them to work.

If we introduce `max_id`, I think we should remove `offset_id`. And when making the request, continue using `offset_id` internally unless `min_id` was also set.

Regarding `prev`, I'm not sure it's a good idea to provide both at the same time. If anything, there should be some sort of `.reversed()`, kind of like in Telethon. But we also need to be very careful here, to make sure it works in all edge cases. And it's more maintenance burden.

aparo:
I'm +1 for the changes. I implemented a downloader and I also had to implement these methods to be able to efficiently process the ranges. offset_id is not enought if you need to evaluate if you have gap in the messages.

Lonami:
> offset_id is not enought if you need to evaluate if you have gap in the messages

Can you help me understand why? A practical example perhaps.

aparo:
I need to check that all the message from a chat are downloaded. I start from offset 0 to x. 
I keep track of the downloaded message number.
Because messages can be deleted or there can be gao issues, I don’t have message for example from 123 to 135. 
the most perfomant check is to ask messages between 123 and 135 and che if the result is 0 means that I have all the messages. The query will no fetch messages if they are missing so it’s performing. Otherwise I should fetch a range and then skip all of them if not in range.

Lonami:
Right, so the usecase is "going back" to a "failed gap", and `offset_id` + `limit` is not desirable because that might needlessly include messages outside of the expected range.

In that case, I think the implementation would make more sense if it had a single method that took a `Range`. Then the user can choose all the combinations, and the questions of "is this inclusive or exclusive" would be obvious. The library can then also take care of using `offset_id` or `min_id/max_id` to correctly fit the usecase.

PR #307, opened 2025-01-24
fix(client): handle errors when editing inline bot messages

AndrielFR:
Fixes the error `DC_ID_INVALID` that occurs when uploading a `tl::enums::InputMedia::PhotoExternal(_)` or `tl::enums::InputMedia::DocumentExternal(_)`.

### DIFF
--- a/lib/grammers-client/src/client/bots.rs
+++ b/lib/grammers-client/src/client/bots.rs
@@ -166,7 +166,13 @@ impl Client {
     ) -> Result<bool, InvocationError> {
         let message: InputMessage = input_message.into();
         let entities = parse_mention_entities(self, message.entities);
-        let result = if message.media.is_some() {
+        if message.media.as_ref().is_some_and(|media| {
+            !matches!(
+                media,
+                tl::enums::InputMedia::PhotoExternal(_)
+                    | tl::enums::InputMedia::DocumentExternal(_),
+            )
+        }) {
             let dc_id = message_id.dc_id();
             self.invoke_in_dc(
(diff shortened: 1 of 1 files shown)


Lonami:
> sometimes appears

Is there no better way to determine when that "sometimes" is?

I'm not entirely happy with the approach here because it is probably being unnecessarily wasteful, and we're doing a catch-all error, which means we may be incorrectly trying to do a second edit.

I haven't heard complaints on this failing from Telethon, and I don't remember it checking for errors like this.

AndrielFR:
> > sometimes appears
> 
> Is there no better way to determine when that "sometimes" is?
> 
> I'm not entirely happy with the approach here because it is probably being unnecessarily wasteful, and we're doing a catch-all error, which means we may be incorrectly trying to do a second edit.
> 
> I haven't heard complaints on this failing from Telethon, and I don't remember it checking for errors like this.

I dont know, sometimes it works just fine with media, othertimes no, i'll investigate, i guess its related to webpage media, not sure

AndrielFR:
> > sometimes appears
> 
> Is there no better way to determine when that "sometimes" is?
> 
> I'm not entirely happy with the approach here because it is probably being unnecessarily wasteful, and we're doing a catch-all error, which means we may be incorrectly trying to do a second edit.
> 
> I haven't heard complaints on this failing from Telethon, and I don't remember it checking for errors like this.

Ok, it only occurs with `InputMedia::PhotoExternal(_)` and `InputMedia::DocumentExternal(_)`.

PR #308, opened 2025-01-26
feat(html): support collapsed attribute in blockquote tag

AndrielFR:
With this, now we can use:
```rust
InputMessage::html("<blockquote expandable>Lorem ip...</blockquote>")
```

### DIFF
--- a/lib/grammers-client/src/parsers/html.rs
+++ b/lib/grammers-client/src/parsers/html.rs
@@ -59,11 +59,13 @@ pub fn parse_html_message(message: &str) -> (String, Vec<tl::enums::MessageEntit
                         entities.push(tl::types::MessageEntityUnderline { offset, length }.into());
                     }
                     tag!("blockquote") => {
+                        let collapsed = attrs.into_iter().any(|a| &a.name.local == "collapsed");
+
                         entities.push(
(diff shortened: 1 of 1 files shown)


AndrielFR:
As for `markdown` equivalent, i don't what to use.

Lonami:
I don't think this is standard HTML right? So I worry about discoverability.

AndrielFR:
> I don't think this is standard HTML right? So I worry about discoverability.

No, i haven't found an enquivalent on HTML

AndrielFR:
> > I don't think this is standard HTML right? So I worry about discoverability.
> 
> No, i haven't found an equivalent on HTML

Ok, bot api uses `expandable` in HTML.
Ref: https://core.telegram.org/bots/api#html-style

As for markdown version, it uses:
```
**> ...
> ...
> ...||
```
Ref: https://core.telegram.org/bots/api#markdownv2-style

Lonami:
While neither are standard, I think it's worthwhile being consistent, for discoverability purposes.

AndrielFR:
> While neither are standard, I think it's worthwhile being consistent, for discoverability purposes.

Ok, so i'll follow bot api

PR #311, opened 2025-02-11
refactored `grammers-tl-gen` to generate each of the types, functions and enums to a separate file

YouKnow-sys:
this PR will modify the tl-gen so it generate each of types, functions and enums to a separate file, the reason I did this is, if we have all of them in a single file the resulted generated file will be so big and that will cause some problems with RA.

I did found some strange problems while doing these changes, like if we add doc comments (`//!`) to start of the generated rust source and then include that file in another file, the rust compiler will give us a error and complain that we can't use these comments here... that's why I no longer add the license and top doc comment directly to the source file when generating, instead I manually added them to the actual module that generated sources will be included in.

I also fixed the small bug after the recent `User` type changes.

### DIFF
--- a/lib/grammers-session/build.rs
+++ b/lib/grammers-session/build.rs
@@ -5,19 +5,23 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-use grammers_tl_gen::{generate_rust_code, Config};
+use grammers_tl_gen::{generate_rust_code, Config, Outputs};
 use grammers_tl_parser::parse_tl_file;
 use std::env;
 use std::fs::File;
-use std::io::{BufWriter, Write};
+use std::io::BufWriter;
 use std::path::Path;
 
 const CURRENT_VERSION: i32 = 3;
 
 fn main() -> std::io::Result<()> {
-    let mut file = BufWriter::new(File::create(
-        Path::new(&env::var("OUT_DIR").unwrap()).join("generated.rs"),
-    )?);
+    let output_dir = Path::new(&env::var("OUT_DIR").unwrap()).to_path_buf();
+    let mut outputs = Outputs {
+        common: BufWriter::new(File::create(output_dir.join("generated_common.rs"))?),
+        types: BufWriter::new(File::create(output_dir.join("generated_types.rs"))?),
+        functions: BufWriter::new(File::create(output_dir.join("generated_functions.rs"))?),
+        enums: BufWriter::new(File::create(output_dir.join("generated_enums.rs"))?),
+    };
 
     // Using boxed variants in the definitions so that deserialization fails if any constructor ID changes.
     let definitions = parse_tl_file(
@@ -37,8 +41,9 @@ fn main() -> std::io::Result<()> {
         ..Default::default()
     };
 
-    generate_rust_code(&mut file, &definitions, CURRENT_VERSION, &config)?;
-    file.flush()?;
+    generate_rust_code(&mut outputs, &definitions, CURRENT_VERSION, &config)?;
+
+    outputs.flush()?;
 
     Ok(())
 }
--- a/lib/grammers-session/src/generated/enums.rs
+++ b/lib/grammers-session/src/generated/enums.rs
@@ -1,0 +1,18 @@
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![allow(clippy::large_enum_variant)]
+
+//! This module contains all of the boxed types, each
+//! represented by a `enum`. All of them implement
+//! [`Serializable`] and [`Deserializable`].
+//!
+//! [`Serializable`]: /grammers_tl_types/trait.Serializable.html
+//! [`Deserializable`]: /grammers_tl_types/trait.Deserializable.html
+
+include!(concat!(env!("OUT_DIR"), "/generated_enums.rs"));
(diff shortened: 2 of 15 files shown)


lib/grammers-client/src/types/chat/mod.rs
```diff
@@ -107,13 +107,17 @@ impl Chat {
         match packed.ty {
             PackedType::User => {
                 let mut user = User::from_raw(tl::types::UserEmpty { id: packed.id }.into());
-                user.raw.access_hash = packed.access_hash;
+                if let tl::enums::User::User(user) = &mut user.raw {
+                    user.access_hash = packed.access_hash;
+                }
```
Lonami, line 112: This would've not worked. I just pushed the proper fix. (And is also why PRs should focus on a single change. Or at least commits should.)

lib/grammers-mtsender/src/utils.rs
```diff
@@ -72,7 +72,7 @@ impl Timeout {
 pub async fn sleep(duration: Duration) {
     #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
     {
-        return tokio::time::sleep(duration).await;
+        tokio::time::sleep(duration).await;
```
Lonami, line 75: Can you explain this change? And, if it needs to stay, please make it a new commit.

lib/grammers-session/build.rs
```diff
@@ -37,8 +46,20 @@ fn main() -> std::io::Result<()> {
...
-    file.flush()?;
+    generate_rust_code(
+        &mut common_file,
+        &mut types_file,
+        &mut functions_file,
+        &mut enums_file,
+        &definitions,
+        CURRENT_VERSION,
+        &config,
+    )?;
```
Lonami, line 57: Way too many parameters of the same type now (4 files in a row). Perhaps we should construct a struct to make sure we don't accidentally write to the wrong file during a refactor.

lib/grammers-session/src/generated/enums.rs
```diff
@@ -0,0 +1,18 @@
...
+//! [`Serializable`]: /grammers_tl_types/trait.Serializable.html
+//! [`Deserializable`]: /grammers_tl_types/trait.Deserializable.html
+
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
```
Lonami, line 16: I'd expect the copyright to be at the top of the file.

lib/grammers-session/src/generated/types.rs
```diff
@@ -0,0 +1,23 @@
...
+//! [`Serializable`]: ../trait.Serializable.html
+//! [`Deserializable`]: ../trait.Deserializable.html
+
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
```
Lonami, line 21: Same here.

lib/grammers-session/src/lib.rs
```diff
@@ -169,7 +169,7 @@ impl Session {
     }
 
     pub fn get_dcs(&self) -> Vec<enums::DataCenter> {
-        self.session.lock().unwrap().dcs.iter().cloned().collect()
+        self.session.lock().unwrap().dcs.to_vec()
```
Lonami, line 172: Assuming this was clippy. Can it be in a different commit?

lib/grammers-tl-types/src/generated/enums.rs
```diff
@@ -0,0 +1,18 @@
...
+//! [`Serializable`]: /grammers_tl_types/trait.Serializable.html
+//! [`Deserializable`]: /grammers_tl_types/trait.Deserializable.html
+
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
```
Lonami, line 16: Same here.

lib/grammers-tl-types/src/generated/functions.rs
```diff
@@ -0,0 +1,27 @@
...
+//! [`Serializable`]: ../trait.Serializable.html
+//! [`RemoteCall`]: trait.RemoteCall.html
+
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
```
Lonami, line 25: Same here.

lib/grammers-tl-types/src/generated/types.rs
```diff
@@ -0,0 +1,23 @@
...
+//! [`Serializable`]: ../trait.Serializable.html
+//! [`Deserializable`]: ../trait.Deserializable.html
+
+// Copyright 2020 - developers of the `grammers` project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
```
Lonami, line 21: Same here.

lib/grammers-mtsender/src/utils.rs
Lonami: Ah right, the cfg confused me, but yeah the parent had no return type anyway.

PR #316, opened 2025-03-25
Change BotInlineResult's raw to public

tholoo:
Currently there is no way to actually use the content retrieved via `client.inline_query()`, Since only the title and query id are accessible.

This PR adds the ability to get the underlying `tl::enums::BotInlineResult` to be able to access fields like `BotInlineMediaResult.photo` from the retrieved query.


Lonami (changes requested):
I think we should rename `result` to `raw` and make it `pub` to make it consistent with all other types.

Lonami (approved):

PR #323, opened 2025-05-02
add method request_qr_code

Cabbache:


lib/grammers-client/src/client/auth.rs
```diff
@@ -258,6 +261,40 @@ impl Client {
         })
     }
 
+    //Returns Some(token) if not logged in, None if logged in
```
Lonami, line 264: This is a normal comment, not a doc-comment.

Cabbache:
This is still in progress, I will submit an update soon.

PR #327, opened 2025-05-23
Don't override an existing DC, when inserting a new one

aav:
- rename the parameter from `dc` to `new_dc` in the `Session::insert_dc`.  
- use the new parameter `new_dc` when checking for an existing data center with the same id, ensuring the method behaves as intended.


Lonami:
Thanks!

PR #337, opened 2025-06-19
Shortcut "while handling diff" return fix

aav:


lib/grammers-session/src/message_box/tests.rs
```diff
@@ -468,6 +468,20 @@ fn test_process_socket_updates_flow_common_already_applied() {
...
 }
 
+#[test]
+fn test_process_socket_updates_flow_common_difference_ok() {
+    reset_time();
+    let mut message_boxes = MessageBoxes::new();
+
+    message_boxes.set_state(state(12, 34, 56, 78));
+    message_boxes.try_begin_get_diff(super::Key::Common);
+
```
Lonami, line 478: We should make sure the call above triggered a difference. Otherwise we might not hit the expected path.

```suggestion

    assert_eq!(
        message_boxes.get_difference(),
        Some(get_difference(12, 56, 78))
    );
```

PR #340, opened 2025-07-22
Update layer to 205

ProDALOR:
Simple update...

https://github.com/tdlib/td/blob/master/td/generate/scheme/telegram_api.tl


Lonami (commented):
Sweet thanks.

PR #347, opened 2025-08-14
Avoid panic in Message::peer_id() when peer_id is missing

Jackhammer88:
This PR changes Message::peer_id() to return Option/Result instead of unwrapping, preventing panics when processing updates with empty messages that lack a peer_id.
Added a new InvocationError::MissingPeerId to propagate this case to callers, and updated related methods (chat(), sender(), etc.) to handle the new return type.


Lonami (approved):
Thanks.

PR #352, opened 2025-09-13
Made session storage more abstract

0xENCRYPTED:
I found current implementation of session storage some kind of ugly because of breaking SOLID principles. 

### Changes made:
- Created `SessionProvider` trait, providing some methods for the storage to implement
- Introduced `snafu` dependency to keep occurring errors more verbose
- Implemented `FileSessionStorage` and `StringSessionStorage`
- Fixed dependants of the session crate
- Implemented helpers for the session to create appropriate storages

P.S. StringSessionStorage might look strange, but i think it's possible to improve it in the future :)

### DIFF
--- a/lib/grammers-session/src/lib.rs
+++ b/lib/grammers-session/src/lib.rs
@@ -11,78 +11,110 @@
 mod chat;
 mod generated;
 mod message_box;
+pub mod session_storage;
 
 pub use chat::{ChatHashCache, PackedChat, PackedType};
 pub use generated::LAYER as VERSION;
 pub use generated::types::UpdateState;
 pub use generated::types::User;
 use generated::{enums, types};
-use grammers_tl_types::deserialize::Error as DeserializeError;
 pub use message_box::PrematureEndReason;
 pub use message_box::{Gap, MessageBox, MessageBoxes, State, UpdatesLike, peer_from_input_peer};
-use std::fmt;
-use std::fs::{File, OpenOptions};
-use std::io::{self, Read, Seek, Write};
+use std::fmt::{Debug, Display};
+use std::io::{Read, Seek, Write};
 use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::path::Path;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
+use snafu::ResultExt;
+
+use crate::session_storage::{file_storage::{FileSessionStorage, FileSessionStorageOptions}, string_storage, SessionProviderError};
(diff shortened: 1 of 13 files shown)


lib/grammers-session/src/lib.rs
```diff
@@ -11,78 +11,110 @@
...
-        File::open(path.as_ref())?.read_to_end(&mut data)?;
+    pub fn load_from_file_or_create<T: AsRef<Path> + 'static>(path: T) -> Result<Self, SessionProviderError> {
+        let fs_storage = FileSessionStorage::default_with_path(path)?;
+
+        match fs_storage.load() {
+            Ok(_) => (),
+            Err(SessionProviderError::InvalidFormat {source}) => {
+                let file_ref = fs_storage.get_file();
+                let mut file = file_ref.lock().unwrap();
+                let size = file.read_to_end(&mut vec![]).unwrap();
```
Lonami, line 65: Probably don't want this unwrap.

lib/grammers-session/src/lib.rs
```diff
@@ -11,78 +11,110 @@
...
+            mode: OpenMode::Open,
+            session: Arc::new(Mutex::new(types::Session{
+                dcs: vec![],
+                user: None,
+                state: None,
+            }))
+        };
+
+        let fs_storage = FileSessionStorage::new(fs_storage_options)?;
+        fs_storage.load()?;
```
Lonami, line 95: This looks like a really strange way of using the API. I'd expect the load to return Self (and not need new).

lib/grammers-session/src/session_storage/mod.rs
```diff
@@ -0,0 +1,44 @@
+pub mod file_storage;
+pub mod string_storage;
+
+use std::io;
+use base64::DecodeError;
+use snafu::{Snafu, prelude::*};
+
+pub trait SessionProvider {
```
Lonami, line 8: I think this is entirely backwards. I think `Session` should be the trait and **not** have any `load` or `save`. Those two would (or not!) exist in the specific implementations, like a `FileSession` or `StringSession`.

Lonami (changes requested):
Not a complete review but, while I appreciate the PR, I have major concerns that require an entirely new approach.

lib/grammers-session/src/session_storage/mod.rs
0xENCRYPTED: For some reason I haven't even thought about it. I will try to mess around with it.

PR #354, opened 2025-09-20
Refactor `InputMessage` and `InputMedia`

### LINKED ISSUE #353: Modify InputMessage.text
My use-case is editing incoming messages to replace some pre-defined keywords. My solution is that when receive a NewMessage update, clone the message then `.into() as InputMessage` then call `editMessage` with it, but currently there is no way to modify the `text` field inside a `InputMessage`, so I have no way to modify just the text.

Do you think we should add a `mut self` version method for text?

SpriteOvO:
Fixes #353.

### Explanation for commits

- Impl `From<Message>` instead of `From<&Message>` for `InputMessage`

  In Rust naming conventions, `from` and `into` more imply the conversion of owned values. If users own a `Message`, they should be able to directly write `msg.into() as InputMessage` instead of `(&msg).into() as InputMessage`; if users have a `Message` reference, they should explicitly clone it then call `into`, like `msg_ref.clone().into()`.

- Change `InputMessage::{text,markdown,html}` to take `self`
- Change `InputMedia::{caption,markdown,html}` to take `self`

  Mentioned in https://github.com/Lonami/grammers/issues/353#issuecomment-3305998409.

- Extend `From<{&str,String}>` to `From<Into<String>>` for `InputMessage`

  Allow more string types (e.g. `Cow<str>, &mut str`, etc.) to be accepted.

### DIFF
--- a/lib/grammers-client/src/types/input_message.rs
+++ b/lib/grammers-client/src/types/input_message.rs
@@ -409,18 +409,12 @@ impl InputMessage {
     }
 }
 
-impl From<&str> for InputMessage {
-    fn from(text: &str) -> Self {
+impl<S> From<S> for InputMessage
+where
+    S: Into<String>,
+{
+    fn from(text: S) -> Self {
         Self::new().text(text)
-    }
-}
-
-impl From<String> for InputMessage {
-    fn from(text: String) -> Self {
-        Self {
-            text,
-            ..Self::default()
-        }
     }
 }
 


Lonami:
> they should explicitly clone

Reason for not doing this is because Messages can be pretty big. That said, I'm not sure how common it is to keep using the message after converting it, so this should be fine.

lib/grammers-client/src/types/input_media.rs
```diff
@@ -21,12 +21,73 @@ pub struct InputMedia {
...
+    /// Creates a new empty media message for input.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Replaces the plaintext in the message.
+    ///
+    /// <div class="warning">
+    /// Note that this method does not modify formatting entities, which may break
+    /// formatting or cause out-of-bounds errors if entities does not match the given
```
Lonami, line 33:
```suggestion
    /// formatting or cause out-of-bounds errors if entities do not match the given
```

lib/grammers-client/src/types/input_media.rs
```diff
@@ -21,12 +21,73 @@ pub struct InputMedia {
...
+    /// text.
+    /// <br><br>
+    ///
+    /// The caller must ensure that formatting entities remain valid for the given text.
+    /// If you need to update formatting entities, call method
+    /// [`InputMedia::fmt_entities`].
```
Lonami, line 39:
```suggestion
    /// If you need to update formatting entities, call [`InputMedia::fmt_entities`].
```

lib/grammers-client/src/types/input_message.rs
```diff
@@ -46,12 +51,68 @@ impl InputMessage {
         self
     }
 
+    /// Replaces the plaintext in the message.
+    ///
+    /// <div class="warning">
+    /// Note that this method does not modify formatting entities, which may break
+    /// formatting or cause out-of-bounds errors if entities does not match the given
```
Lonami, line 58:
```suggestion
    /// formatting or cause out-of-bounds errors if entities do not match the given
```

lib/grammers-client/src/types/input_message.rs
```diff
@@ -46,12 +51,68 @@ impl InputMessage {
...
+    /// text.
+    /// <br><br>
+    ///
+    /// The caller must ensure that formatting entities remain valid for the given text.
+    /// If you need to update formatting entities, call method
+    /// [`InputMessage::fmt_entities`].
```
Lonami, line 64:
```suggestion
    /// If you need to update formatting entities, call [`InputMessage::fmt_entities`].
```

Lonami, line 64: I also think this does not need to be part of the warning (for both here and the media).

SpriteOvO:
> > they should explicitly clone
> 
> Reason for not doing this is because Messages can be pretty big. That said, I'm not sure how common it is to keep using the message after converting it, so this should be fine.

This isn't a big deal, we can always add a method like this in the future to construct `InputMessage` by a reference, if needed.

```rust
fn to_input_message(&self) -> InputMessage { /* ... */ }
```

lib/grammers-client/src/types/input_media.rs
SpriteOvO: Fixed

lib/grammers-client/src/types/input_message.rs
SpriteOvO: Fixed

SpriteOvO: Fixed

PR #360, opened 2025-10-21
Refactor client to be network-agnostic

Lonami:
`grammers-client` no longer concerns itself with stepping. This is now done in the `grammers-mtsender` crate via the new `SenderPool`. The `Client` now only needs a way to `invoke_in_dc` (this may be possible to turn into a `trait` in the future so users can provide their own `SenderPool`-esque logic).

The refactor is not fully complete:

- [x] Review that dead connections notify pending requests correctly.
- [x] Review that failing to connect does not panic.
- [ ] Lost reconnection policies.
- [x] Lost proxy support.
- [ ] Lost WASM support.
- [x] Lost authentication exports (and thus downloads from other DCs).
- [ ] Likely want a way to terminate idle connections (like Telethon).
- [x] Improve back some logs to re-include the constructor name.
- [ ] Should be possible to reduce cloning when re-attempting requests.
- [x] The way the handle and particularly the updates are dealt with is somewhat clunky.
- [x] Duplicating the `api_id` is unfortunate (future refactor: move `api_hash` to just the methods that need it? Or is it nice to give early?)
- [x] Probably rename client's `Config` to `Configuration` to be consistent.
- [ ] Some other things I might be missing.
- [x] Quit is nice but query / reconfigure feel like a hack. Should probably just share the session. And Quit should not be a "forget and you're locked" - if all handles are dropped, the pool should die out.

### DIFF
--- a/lib/grammers-client/src/client/messages.rs
+++ b/lib/grammers-client/src/client/messages.rs
@@ -86,7 +86,7 @@ fn map_random_ids_to_messages(
 }
 
 pub(crate) fn parse_mention_entities(
-    _client: &Client,
+    client: &Client,
     mut entities: Vec<tl::enums::MessageEntity>,
 ) -> Option<Vec<tl::enums::MessageEntity>> {
     if entities.is_empty() {
@@ -104,7 +104,12 @@ pub(crate) fn parse_mention_entities(
                     length: mention_name.length,
                     user_id: tl::enums::InputUser::User(tl::types::InputUser {
                         user_id: mention_name.user_id,
-                        access_hash: 0,
+                        access_hash: client
+                            .0
+                            .session
+                            .peer(grammers_session::PeerRef::User(mention_name.user_id))
+                            .and_then(|peer| peer.hash())
+                            .unwrap_or(0),
                     }),
                 }
                 .into()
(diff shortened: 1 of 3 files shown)


Lonami:
Sorry @er-azh, I really appreciate all your hard work on adding WASM support to grammers, but I feel very overwhelmed trying to work on the library while also attempting to maintain compatibility with a platform that I do not even know how to test (or personally care for at current time). So I'll be yanking out most of it to make it easier on myself to complete the refactor and be sure it works.

There must be a less-intrusive way to run grammers on WASM, but I do not know what that is. Frankly, it might need to be a fork of the `grammers-mtsender`. And perhaps everything relying on time / instants should be dependency-injected somehow (to also aid with testing). Removing `web_time` might be overkill but since again I do not need it yet I prefer to trim down the amount of dependencies.

If continued WASM support is needed, for now, I think it would be best to fork at the point before this PR and perhaps continue updating the layer as needed.

I am hoping that the changes introduced by this PR pave the way forward to make that much simpler. Namely, by hopefully turning the `SenderPool` into something abstract enough that it can be trivially swapped and made work on other platforms. (Personally, I'm hopeful I can also reuse it from Python at some point.)

PR #367, opened 2025-11-11
General improvements

Lonami:


grammers-client/src/types/message.rs
```diff
@@ -231,7 +231,7 @@ impl Message {
         self.raw.id()
     }
 
-    pub(crate) fn peer_ref(&self) -> PeerRef {
+    pub fn peer_ref(&self) -> PeerRef {
```
Lonami: Yeah fair, I wanted to revisit these and make sure they were consistent at some point.

I think we'll want `peer` and `peer_ref` on pretty much everything that contains a peer (and sometimes `sender` and `sender_ref`).

grammers-tl-parser/src/tl/definition.rs
```diff
@@ -146,55 +143,49 @@ impl FromStr for Definition {
                     None
                 }
 
-                // If the parameter is a flag definition save both
-                // the definition and the parameter.
-                Ok(Parameter {
```
Lonami: I can never remember the syntax order but maybe we could've used `@ p`.

This is fine too though, just more nesting.

grammers-session/src/peer.rs
```diff
@@ -255,138 +255,227 @@ impl fmt::Display for PeerId {
...
 }
 
 impl From<tl::types::PeerChannel> for PeerId {
+    #[inline]
     fn from(channel: tl::types::PeerChannel) -> Self {
-        PeerId::channel(channel.channel_id)
+        <Self as From<&tl::types::PeerChannel>>::from(&channel)
+    }
+}
+impl<'a> From<&'a tl::types::PeerChannel> for PeerId {
```
Lonami: These are nice to have (and I can see why people would use macros), but it's again unfortunate this only works for one reference type.

grammers-client/src/types/peer/channel.rs
```diff
@@ -117,6 +117,15 @@ impl Channel {
             .unwrap_or_default()
     }
 
+    /// Additional information about this channel.
+    #[inline]
+    pub fn kind(&self) -> Option<ChannelKind> {
+        match <ChannelKind as TryFrom<&Channel>>::try_from(self) {
+            Ok(channel_kind) => Some(channel_kind),
+            Err(()) => None,
+        }
```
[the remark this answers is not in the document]
bb010g: This was intentional, to catch changes to the `Err` value. `<Result<_, ()>>::ok(<ChannelKind as TryFrom<&Channel>>::try_from(self))` could work.

grammers-tl-parser/src/tl/definition.rs
bb010g: I tried `p @ …` first; I couldn't resolve the resulting mutable reference issues.

grammers-session/src/peer.rs
bb010g: What other types are you referring to by “only one reference type”?

grammers-client/src/types/peer/channel.rs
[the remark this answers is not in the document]
Lonami: IMO there's no need to catch changes to `Err` here. Changing the return type would be breaking anyway.

grammers-session/src/peer.rs
bb010g: That's what [`Deref`](<https://doc.rust-lang.org/core/ops/trait.Deref.html>) is for.

PR #393, opened 2026-01-14
Deny potential integer overflows

dimentyy:
In short, use `i32::try_from(some_usize_var).unwrap()` instead of `some_usize_var as i32` to avoid potential integer overflow in `grammers-crypto`, `grammers-mtproto` and `grammers-mtender`.

Added `#[deny(clippy::cast_possible_truncation, clippy::cast_sign_loss)]` to enforce these rules and `#[expect]` for statements where integer overflow is acceptable or impossible.

It is much better to panic than to continue with bad data.


dimentyy:
All three transports (Abridged, Intermediate and Full) were tested

PR #405, opened 2026-02-09
Add SOCKS4 and HTTP CONNECT proxy support

ooopus:


Lonami (changes requested):
Please submit separate PRs for different features. I'm OK with merging the addition of `requires_join_request`, but manually implementing SOCKS4 is not something I'm comfortable with (I do not want to maintain or debug that part of the code). I would rather add an extra optional dependency to take care of it.

Lonami:
Closing in favor of #408.

PR #406, opened 2026-02-09
IPv6 Support

tsukinaha:
We may need a more graceful implement, such as a fallback or something else.  
I would like some guidance on what the behavior it should look like.

Currently, setting the `PREFER_V6` env var simply forces IPv6 connections.

### DIFF
--- a/grammers-mtsender/src/sender_pool.rs
+++ b/grammers-mtsender/src/sender_pool.rs
@@ -292,7 +292,6 @@ impl SenderPoolRunner {
         let transport = transport::Full::new;
 
         let address = if self.connection_params.use_ipv6 {
-            log::info!("Using IPv6 connection");
             dc_option.ipv6.into()
         } else {
             dc_option.ipv4.into()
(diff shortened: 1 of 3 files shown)


grammers-mtsender/src/sender_pool.rs
```diff
@@ -291,22 +291,30 @@ impl SenderPoolRunner {
     ) -> Result<Sender<transport::Full, mtp::Encrypted>, InvocationError> {
         let transport = transport::Full::new;
 
+        let address = match std::env::var("PREFER_V6") {
```
Lonami: Should be part of the `ConnectionParams` instead. Maybe a new `use_ipv6` boolean flag.

Lonami: I don't understand. You can still use an environment variable in your application that depends on grammers to set the `ConnectionParam` when creating the `SenderPool`.

tsukinaha: Ah, I see what you mean. 
Then I will do it that way.

PR #407, opened 2026-02-09
Add requires_join_request methods to Channel and Group

ooopus:
- Add `requires_join_request()` method to `Channel` that returns `self.raw.join_request`
- Add `requires_join_request()` method to `Group` that checks the underlying `Chat` enum variant


Lonami:
Thanks.

